PR #2, opened 2016-10-13
Server refactor

nathf:
* Use only one entry point for both server and client instead of previous two.
* Remove `widgets` dir
* Ability to specify port from ENV variable

These features really need documentation


nathf:
Agree - will add it to the backlog

PR #4, opened 2016-10-16
Feature/case study view

nathf:


src/bundles/CaseStudy/components/View/View.json
```diff
@@ -0,0 +1,24 @@
...
+    { "title": "Outcome", "items": [
```
matt-sm: In a section, whats the difference between content and items?

nathf: Items are lists
Content is just a word blob

matt-sm:
As we discussed, would be good to see case study view rendering via npm start

PR #6, opened 2016-10-17
Move server to webpack

nathf:
Key things to note in this PR

* `npm run build` must be run before starting the server.
* `src` is assumed when calling the rendering service, you can now just call `bundles/CaseStudy/CaseStudyViewWidget.js` instead of `src/bundles/CaseStudy/CaseStudyViewWidget.js`


matt-sm:
How can I test this locally?

nathf:
With these two:
`npm run build`
`npm run server`

PR #19, opened 2016-10-24
Seller registration - start page

matt-sm:
Simple start page copied from the prototype

Deed + start url passed as props

### DIFF
--- a/src/bundles/SellerRegistration/StartWidget.js
+++ b/src/bundles/SellerRegistration/StartWidget.js
@@ -1,0 +1,27 @@
+import React from 'react'
+import RegisterComponent from '../../RegisterComponent'
+
+export const StartWidget = (props) => {
+    return (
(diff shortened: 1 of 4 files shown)


src/bundles/SellerRegistration/StartWidget.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import RegisterComponent from '../../RegisterComponent'
+
+export const StartWidget = (props) => {
+    return (
```
nathf: Since this is a presentational component, you can simplify this.

from:
```javascript
export const StartWidget = (props) => {
  return (
    <div>
      ...
    </div>
  )
}
```
to:
```javascript
export const StartWidget = (props) => (
  <div>
    ...
  </div>
)
```

This is fine for now, but this might be helpful in the future.

PR #21, opened 2016-10-24
Add integrated RadioList control

nathf:


src/shared/form/RadioList.js
```diff
@@ -0,0 +1,67 @@
...
+          return (
+            <div key={i}>
+              <Control.radio
+                model={model}
+                name={name}
+                id={fieldId}
+                value={option.value}
+                validators={validators}
+              />
+              <label htmlFor={fieldId}>{option.label}</label>
```
elisechant: You could wrap the control in label here to make the label (explicit) rather than have it aside (implicit). This might improve tabbing, highlight or clicking.

src/shared/form/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+Consisting of:
+
+| Prop          | Type                    |
+|---------------|-------------------------|
+|  id           |  String (required)      |
+|  name         |  String (required)      |
+|  htmlFor      |  String (required)      |
+|  label        |  String (required)      |
+|  model        |  String|func (required) |
```
elisechant: when would a model be a func?

src/shared/form/RadioList.js
```diff
@@ -0,0 +1,67 @@
...
+  } = props;
+  return (
+    <div className="field">
+      <fieldset>
+        <legend>{label}</legend>
+        <Errors
+          model={model}
+          show="touched"
+          messages={messages}
+          wrapper={(props) => (
```
elisechant: which props?

src/shared/form/RadioList.js
```diff
@@ -0,0 +1,67 @@
...
+  ]).isRequired,
+  options: React.PropTypes.arrayOf(React.PropTypes.shape({
+    label: React.PropTypes.string,
+    value: React.PropTypes.oneOfType([
+      React.PropTypes.string,
+      React.PropTypes.number
+    ])
+  })).isRequired,
+
+  validators: React.PropTypes.object,
```
nathf: Indeed

src/shared/form/README.md
nathf: Its an accepted type of the library, just supporting the flow through.

src/shared/form/RadioList.js
nathf: Also would make screen readers read these controls properly. Need to double check the styling side effects of this change, otherwise I'm all for it

matt-sm:
Closing for now

PR #23, opened 2016-10-25
Form accessibility

nathf:


src/bundles/CaseStudy/redux/modules/errorMessage.js
```diff
@@ -0,0 +1,21 @@
+const MESSAGE = 'error/message'
+const REMOVE = 'error/remove'
+
+export default function reducer(state = {}, action = {}) {
+  const { type, key, value } = action
+  switch (type) {
+    case MESSAGE:
+      state[key] = value;
```
elisechant: does this not break immutability?

nathf: Yep sure does - I've addressed it in next PR.

This whole module is awful, needs to be ripped out/moved

PR #25, opened 2016-10-25
add prod deploy config

maxious:


circle.yml
```diff
@@ -19,3 +19,9 @@ deployment:
     commands:
       - cf login -a https://api.system.staging.digital.gov.au -o dto -u $CF_USER_STAGING -p $CF_PASSWORD_STAGING
       - cf push
+  production:
+    tag: /release-.*/
```
nathf: Shouldn't we use semver versioning instead of this style? 

/cc @matt-sm

matt-sm: yes we could
on other repos we use different tags for different types of builds eg release-, monitor-
at the time I also wanted to ensure the person creating the release knew they were deploying to live, not just tagging a branch

maxious: should we still prefix with "v" or just the raw semver?

PR #30, opened 2016-11-01
start of casestudy pagetwo

nathf:
Not finished, but has quite a few improvements that would be beneficial to other streams.


src/bundles/CaseStudy/redux/modules/reducer.js
```diff
@@ -1,14 +1,15 @@
...
 
 export default combineReducers({
   casestudy,
-  form_options: (state = { mode: 'add' }) => state,
-  form: combineForms({
-    caseStudy: {}
-  }, 'form'),
+  form_options,
   errorMessage,
+  ...createForms({
```
elisechant: ?

nathf: Spread the forms at the root level.

Instead of having a state shape of:

```
{
  form: {
    caseStudy: { ... }
  }
}
```
it can now be

```
{
  caseStudyForm: { ... }
}
```

src/shared/form/ErrorBox.js
```diff
@@ -15,6 +15,11 @@ class ErrorBox extends React.Component {
```
matt-sm: this file still has a reference to CaseStudy - I need to use it in SellerRegistration

src/bundles/CaseStudy/redux/modules/selector.js
```diff
@@ -1,53 +1,68 @@
 import { createSelector } from 'reselect';
```
nathf: Have moved all common redux modules to a common place. Still ways away from fixing the tests though.

PR #36, opened 2016-11-03
Move common functions to BaseForm

nathf:


matt-sm:
you should also refactor YourInfoForm

PR #52, opened 2016-11-11
First pass of fixture testing react features

nathf:


fixtures/case-study.fixture.js
```diff
@@ -0,0 +1,26 @@
...
+      browser.setValue('#approach', 'approach field');
+      browser.setValue('#outcome', 'outcome field');
+
+      // ErrorBox should be gone since all required fields have values.
+      assert.equal(browser.isExisting('.callout--warning'), false);
+
+      browser.click('[type="submit"]');
+
+      // On View case study page - successfully created.
+      assert.equal(browser.getText('h1'), 'Integration Test Case Study');
```
matt-sm: I assume we would need to delete this case study after test completed?

nathf: Yea thats the plan. This test doesn't even create a case study as there is an issue on staging with creation.

wdio.conf.js
```diff
@@ -0,0 +1,264 @@
...
+        {
+            browser: 'internet explorer',
+            project: process.env.CIRCLE_PROJECT_REPONAME,
+            build: process.env.CIRCLE_BUILD_NUM
+        },
+        {
+            browser: 'safari',
+            project: process.env.CIRCLE_PROJECT_REPONAME,
+            build: process.env.CIRCLE_BUILD_NUM
+        }
```
nathf: @matt-sm Do we want to add Mobile platforms in here?

PR #59, opened 2016-11-15
Feature/add new relic

nathf:


circle.yml
```diff
@@ -10,8 +10,9 @@ dependencies:
     - sudo dpkg -i cf-cli_amd64.deb
     - cf -v
   override:
-    - npm install
-    - npm run build
+    - npm i -g yarn@0.17.0
```
elisechant: why 0.17.x instead of 0.16.x?

nathf: Couldn't find a good reason why (since there are no release notes for 0.17). Bumping back to latest stable (0.16.1) that we know of!

elisechant: Oh yawn! 😔

PR #66, opened 2016-11-17
Domain & Pricing components

nathf:


matt-sm:
- I believe the pricing is at the service level, not the domain level
- may be less confusing to create application.pricing[]

matt-sm:
I would also just not save a deleted domain/service in the application.  Eg. 

 "User research & design": {
          "Inclusive design (Accessibility, disabilty & minority groups)": false

Just remove it rather than leaving it there with "false"

matt-sm:
If you remove a domain (role) that has a price, the deleted role's pricing is still posted

PR #69, opened 2016-11-21
Add services and pricing to seller summary

maxious:


matt-sm:
Perhaps I've been reading too many react tutorials lately, but is there an opportunity here to split this into separate components?

PR #79, opened 2016-11-25
Feature/file upload

matt-sm:
- tests to come later


src/bundles/SellerRegistration/components/DocumentsForm/DocumentsForm.js
```diff
@@ -0,0 +1,178 @@
...
+
+  static propTypes = {
+    action: React.PropTypes.string,
+    csrf_token: React.PropTypes.string,
+    form: React.PropTypes.object.isRequired,
+    dispatch: React.PropTypes.func.isRequired,
+    model: React.PropTypes.string.isRequired
+  }
+
+  state = {
```
elisechant: state should be set up in the constructor? this seems like an anti-pattern to me

src/bundles/SellerRegistration/components/DocumentsForm/DocumentsForm.js
```diff
@@ -0,0 +1,178 @@
...
+    { 'label': 'Signed copy of Deed of Standing Offer', 'id': 'deed' },
+  ]
+
+  onUpload(id, e) {
+    e.preventDefault();
+    const { dispatch, model, action, csrf_token } = this.props;
+    const file = this.state[id].file;
+
+    this.setState({
+      [id]: Object.assign({}, this.state[id], {'uploading': true, 'file': void 0}),
```
elisechant: could `null` be better than `void 0`?

src/bundles/SellerRegistration/components/DocumentsForm/DocumentsForm.js
```diff
@@ -0,0 +1,178 @@
...
+
+import formProps     from '../../../../shared/reduxModules/formPropsSelector';
+
+class DocumentsForm extends BaseForm {
+
+  static propTypes = {
+    action: React.PropTypes.string,
+    csrf_token: React.PropTypes.string,
+    form: React.PropTypes.object.isRequired,
+    dispatch: React.PropTypes.func.isRequired,
```
elisechant: shouldn't we be accessing dispatch from connect?

src/bundles/SellerRegistration/components/DocumentsForm/DocumentsForm.js
```diff
@@ -0,0 +1,178 @@
...
+    this.setState({
+      [id]: Object.assign({}, this.state[id], {'uploading': true, 'file': void 0}),
+      errors: Object.assign({}, this.state.errors, {[id]: void 0})
+    })
+
+    dispatch(actions.omit(`${model}.documents`, id))
+
+    let data = new FormData()
+    data.append(id, file)
+    fetch(`${action}/${id}`, {
```
elisechant: i thought that you guys had api available in dispatch to handle fetch src/shared/reduxModules/api.js. Also, this could be an action?

src/bundles/SellerRegistration/components/DocumentsForm/DocumentsForm.js
```diff
@@ -0,0 +1,178 @@
...
+            validators={{
+              documents: (documents) => Object.keys(documents).length === 4
+            }}
+          >
+            {
+              csrf_token && (
+              <input type="hidden" name="csrf_token" id="csrf_token" value={csrf_token} />
+            )}
+
+            {this.files.map((file, i) => {
```
elisechant: read this as an odd naming convention. Interpreted `this.files` as an existing object, but here it is not uploaded yet.

src/bundles/SellerRegistration/components/DocumentsForm/DocumentsForm.js
nathf: They are equivalent

src/bundles/SellerRegistration/components/DocumentsForm/DocumentsForm.js
nathf: Prefer to work with `undefined`s over `null`s

PR #87, opened 2016-11-29
Add more fields to application preview

maxious:
@nathf Is there an easier way to do these templating loops?


src/bundles/SellerRegistration/components/ApplicationPreview/ApplicationPreview.js
```diff
@@ -24,9 +24,54 @@ const ApplicationPreview = ({application}) => (
             </li>
             }
             <li>
+                <p>Case Studies</p>
+                <article>
+                    <ul>
+                        {Object.keys(application.casestudies).map((key, val) =>
```
nathf: If you return like this:

```javascript
.map(() => {
   const assignment = 'foo';
   return (
      // ...
   )
})
```

Will make things a bit easier to read and reference.

PR #96, opened 2016-12-01
Icons

nathf:


package.json
```diff
@@ -43,6 +43,7 @@
     "rimraf": "2.5.4",
     "strip-ansi": "3.0.1",
     "style-loader": "0.13.1",
+    "svg-inline-loader": "0.7.1",
```
matt-sm: noticed that package is no longer actively developed

PR #97, opened 2016-12-01
Move reference to case study

matt-sm:


src/bundles/CaseStudy/components/CaseStudyForm/CaseStudyForm.js
```diff
@@ -173,6 +174,49 @@ class CaseStudyForm extends BaseForm {
               validators={{ validLinks }}
             />
 
+            <h3>Referee</h3>
+
+            <Textfield
+                model={`${model}.refereeName`}
```
nathf: these model attributes should be snake cased

src/bundles/SellerRegistration/redux/modules/reducer-signup.js
```diff
@@ -56,9 +53,6 @@ export default combineReducers({
     },
```
nathf: You will need to add the reference fields down here too, this is important in the event of a dispatched reset.

PR #101, opened 2016-12-05
Minor tweaks

matt-sm:
- make website mandatory
- persist additional info
- change additional_info to seller_type


src/bundles/SellerRegistration/redux/modules/reducer-signup.js
```diff
@@ -33,7 +33,8 @@ export default combineReducers({
         suburb: '',
         state: '',
         postal_code: ''
-      }
+      },
+      additional_info: {}
```
nathf: This is changing to `seller_type` I believe

PR #106, opened 2016-12-06
Start page copy

matt-sm:


src/bundles/SellerRegistration/components/Start.js
```diff
@@ -23,13 +27,11 @@ const Start = ({ signup, deed, onClick }) => (
 
 Start.defaultProps = {
   onClick: () => {},
-  deed: '#',
-  signup: '#'
+  criteria: '#'
```
maxious: this doesn't need to be configurable, right?

PR #111, opened 2016-12-07
Copy changes

maxious:


src/bundles/CaseStudy/components/CaseStudyForm/__snapshots__/CaseStudyForm.snaps.test.js.snap
```diff
@@ -32,6 +32,7 @@ exports[`test CaseStudyForm renders 1`] = `
           aria-describedby={undefined}
           className=""
           id="title"
+          maxLength={undefined}
```
maxious: One of the changes was adding to the postcode field a 4 character limit. Which means every textfield on every screen needs the ability to be passed a maxLength.

matt-sm: ok

PR #114, opened 2016-12-08
Profile enhancements

maxious:


src/shared/SellerProfile/Body.js
```diff
@@ -83,22 +92,31 @@ const Body = (props) => {
...
           <span>{address.state} {address.postal_code}</span>
         </p>
-
+          { interstate && <p><b>Able to work interstate</b></p>}
         <b>ABN</b><br/>
         <p>{abn}</p>
       </Row>
+      <Row title="Documents" show={Object.keys(documents).length}>
+          {Object.keys(documents).map((key, val) =>
+              <p key={val}><b>{key}:</b> <a href={`../../documents/${documents[key]}`}>{ documents[key] }</a></p>
```
matt-sm: the application id is missing from the path when I test this locallly

src/shared/SellerProfile/Body.js
```diff
@@ -66,46 +75,90 @@ const Body = (props) => {
...
-
+          { interstate && <p><b>This seller is able work interstate</b></p>}
         <b>ABN</b><br/>
         <p>{abn}</p>
       </Row>
+      <Row title="Documents" show={Object.keys(documents).length}>
+          {Object.keys(documents).map((key, val) =>
+              <p key={val}>
+
+                  <a href={`${documentsUrl}${documents[key]}`}><Switch>
```
matt-sm: could have done this without switch eg.  have another file that maps key to string
{
   financial: 'Financial Statement'
}

maxious: Made an object to hold these mappings instead of using switch.

PR #133, opened 2017-01-02
Add a new badge description and remove some data from seller type

maxious:


maxious:
Existing data will need to be migrated by hand.

PR #135, opened 2017-01-03
File expiry format

nathf:


src/shared/form/Datefield.js
```diff
@@ -1,60 +1,100 @@
 import React from 'react';
-import { Control } from 'react-redux-form';
+import { connect } from 'react-redux';
+import { actions } from 'react-redux-form';
+import format from 'date-fns/format';
+import get from 'lodash/get';
```
elisechant: is this a performance boost? could be `import {get} from 'lodash'`

src/shared/form/Datefield.js
```diff
@@ -1,60 +1,100 @@
...
-    />
-  </span>
-);
-
-Datefield.defaultProps = {
-    type: "text"
+class Datefield extends React.Component {
+
+  state = {
+    day: void 0,
```
elisechant: what is with you guys and `void`? can't it just be empty string or null?

src/shared/form/Datefield.js
```diff
@@ -1,60 +1,100 @@
...
+    month: void 0,
+    year: void 0
+  }
+
+  constructor(props) {
```
elisechant: technically `props, context`

src/shared/form/Datefield.js
```diff
@@ -1,60 +1,100 @@
...
+  onChange(e) {
+    const { model, dispatch } = this.props;
+    this.setState({
+      [e.target.name]: e.target.value
+    }, () => {
+      const { year, month, day } = this.state;
+
+      let date = format(new Date(Number(year), Number(month) - 1, Number(day)), 'YYYY-MM-DD');
+
+      dispatch(actions.change(model, date));
```
elisechant: pass handler in on props

src/shared/form/Datefield.js
nathf: Prefer to work with undefineds over string or nulls

src/shared/form/Datefield.js
nathf: Context as second arg? That would be redundant here anyway since its not used?

config/webpack.config.dev.js
```diff
@@ -89,6 +89,10 @@ var loaders = [
       {
         test: /\.svg$/,
         loader: 'svg-inline'
+      },
+      {
+        test: /\.scss$/,
```
nathf: Not currently using sass in webpack - have to ignore to avoid errors, just a temporary ignore until the styling as a whole is addressed (maybe css-modules)

src/shared/form/Datefield.js
matt-sm: we :heart: `void 0`

config/webpack.config.dev.js
elisechant: k. +1 css-modules

src/shared/form/Datefield.js
elisechant: k

src/shared/form/Datefield.js
elisechant: you guys are mad. lol

src/shared/form/Datefield.js
elisechant: except super might

elisechant:
🚢

PR #143, opened 2017-01-09
Fix/refactor profile

nathf:


src/bundles/SellerRegistration/components/ApplicationPreview/ApplicationPreview.js
```diff
@@ -50,7 +51,7 @@ const mapStateToProps = ({ application }, { documentsUrl, caseStudyUrl, onClick
       linkedin,
       abn,
       address,
-      interstate,
+      travel: interstate,
```
maxious: Does changing this need a change any other code?

nathf: `interstate` is referenced in body, so just aliasing `travel` to that

PR #145, opened 2017-01-10
Connected link

nathf:
Dispatch event on specific link clicks to focus heading on next screen, doesnt feel good.

### DIFF
--- a/src/bundles/CaseStudy/components/DomainList/DomainList.js
+++ b/src/bundles/CaseStudy/components/DomainList/DomainList.js
@@ -11,6 +11,9 @@ import ErrorBox       from '../../../../shared/form/ErrorBox';
 import StatefulError  from '../../../../shared/form/StatefulError';
 import formProps      from '../../../../shared/reduxModules/formPropsSelector';
 
+import ConnectedLink from '../../../SellerRegistration/components/ConnectedLink';
+import { linkClick } from '../../../SellerRegistration/redux/modules/application';
+
 import CaseStudyForm from '../CaseStudyForm';
 import View from '../View';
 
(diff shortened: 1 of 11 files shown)


matt-sm:
You may need to show me why you need to do this

src/bundles/SellerRegistration/components/Finish.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 
 const Finish = ({onClick}) => (
     <div>
-        <h1>That’s it for now!</h1>
+        <h1 tabIndex="-1">That’s it for now!</h1>
```
elisechant: i think you need to describe intent here. maybe the h1 should be wrapped in a hoc from the same file as connectedlink?

src/bundles/SellerRegistration/redux/helpers.js
```diff
@@ -92,4 +92,14 @@ export const findDirtyForms = (state = {}) => {
       valid[key] = true;
       return valid;
     }, {});
+}
+
+// Not the greatest thing, but since we're not mutating anything
+// in the DOM and just managing focus, I think we're okay.
+// findDOMNode doesn't break the lines of encapsulation.
+export const focusHeading = () => {
```
elisechant: yeah this should get wrapped up with your h1 component _ideally_

src/bundles/CaseStudy/components/DomainList/DomainList.js
```diff
@@ -11,6 +11,9 @@ import ErrorBox       from '../../../../shared/form/ErrorBox';
 import StatefulError  from '../../../../shared/form/StatefulError';
 import formProps      from '../../../../shared/reduxModules/formPropsSelector';
 
+import ConnectedLink from '../../../SellerRegistration/components/ConnectedLink';
+import { linkClick } from '../../../SellerRegistration/redux/modules/application';
```
elisechant: i think this view needs a container?

elisechant:
@nathf maybe you could consider a factory pattern for Connected Links as props?

src/bundles/SellerRegistration/components/Finish.js
nathf: This makes a lot of sense to me.

PR #146, opened 2017-01-10
Enable case study per service mandatory validation

maxious:


nathf:
Shouldnt this take into account preselected services?

maxious:
Like legacy domains? Decision is for now to make it mandatory for all.

PR #147, opened 2017-01-10
Add profile edit flow

maxious:


nathf:
A lot of duplication here, whats the difference between this and the original?

maxious:
Hmm, I've added a different implementation using filterSteps which still differentiates the entrypoint if we need to diverge more later.

src/shared/SellerProfile/Body.js
```diff
@@ -120,7 +120,7 @@ const Body = (props) => {
       </Row>
 
       <Row title="Documents" show={!isEmpty(documents)}>
-          {Object.keys(documents).map((key, val) => {
+          {!isEmpty(documents) && Object.keys(documents).map((key, val) => {
```
nathf: Is this required? Shouldn't the show prop on Row cater for this case?

maxious: Yes, without it empty documents was still crashing render. I'm guessing it's still evaluated but passed to Row which then decides if it's not rendered?

nathf: Weird, I'll take a look later

PR #150, opened 2017-01-12
CSS modules

nathf:
First pass at using CSS modules in DMP.

### DIFF
--- a/config/webpack.config.dev.js
+++ b/config/webpack.config.dev.js
@@ -173,7 +173,7 @@ module.exports = [{
       test: /\.css$/,
       loader: ExtractTextPlugin.extract(
         'style', 
-        'css?modules&importLoaders=1&context=' + paths.appSrc + '&localIdentName=[path]___[name]__[local]___[hash:base64:5]!postcss-loader'
+        'css?modules&importLoaders=1&context=' + paths.appSrc + '&localIdentName=[path]___[name]__[local]___[hash:base64:5]!postcss'
       )
     })
   },
--- a/config/webpack.config.dev.js
+++ b/config/webpack.config.dev.js
@@ -51,12 +51,9 @@ var loaders = [
             name: 'react-scripts'
           }),
           plugins: [
-            'transform-react-jsx',
             [
               'react-css-modules',
-              {
-                context: paths.appSrc,
-              }
+              { context: paths.appSrc, }
             ]
           ]
         }
--- a/config/webpack.config.dev.js
+++ b/config/webpack.config.dev.js
@@ -7,6 +7,7 @@ var InterpolateHtmlPlugin = require('react-dev-utils/InterpolateHtmlPlugin');
 var WatchMissingNodeModulesPlugin = require('react-dev-utils/WatchMissingNodeModulesPlugin');
 var getClientEnvironment = require('./env');
 var paths = require('./paths');
+var ExtractTextPlugin = require('extract-text-webpack-plugin');
 
 // Webpack uses `publicPath` to determine where the app is being served from.
 // In development, we always serve from the root. This makes config easier.
@@ -48,17 +49,17 @@ var loaders = [
           // https://github.com/facebookincubator/create-react-app/issues/483
           cacheDirectory: findCacheDir({
             name: 'react-scripts'
-          })
+          }),
+          plugins: [
+            'transform-react-jsx',
+            [
+              'react-css-modules',
+              {
+                context: paths.appSrc,
+              }
+            ]
+          ]
         }
-      },
-      // "postcss" loader applies autoprefixer to our CSS.
-      // "css" loader resolves paths in CSS and adds assets as dependencies.
-      // "style" loader turns CSS into JS modules that inject <style> tags.
-      // In production, we use a plugin to extract that CSS to a file, but
-      // in development "style" loader enables hot editing of CSS.
-      {
-        test: /\.css$/,
-        loader: 'style!css!postcss'
       },
       // JSON is not enabled by default in Webpack but both Node and Browserify
       // allow it implicitly so we also enable it.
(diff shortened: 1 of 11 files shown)


config/webpack.config.dev.js
```diff
@@ -7,6 +7,7 @@ var InterpolateHtmlPlugin = require('react-dev-utils/InterpolateHtmlPlugin');
 var WatchMissingNodeModulesPlugin = require('react-dev-utils/WatchMissingNodeModulesPlugin');
 var getClientEnvironment = require('./env');
 var paths = require('./paths');
+var ExtractTextPlugin = require('extract-text-webpack-plugin');
```
elisechant: no need for "extract" on dev mode, webpack has js first paradigm, css comes from the devserver socket, no need to write to file

config/webpack.config.dev.js
```diff
@@ -48,18 +49,18 @@ var loaders = [
           // https://github.com/facebookincubator/create-react-app/issues/483
           cacheDirectory: findCacheDir({
             name: 'react-scripts'
-          })
+          }),
+          plugins: [
+            'transform-react-jsx',
```
elisechant: what does this do?

config/webpack.config.dev.js
```diff
@@ -171,7 +172,13 @@ module.exports = [{
         include: paths.appSrc,
       }
     ],
-    loaders: loaders
+    loaders: loaders.concat({
+      test: /\.css$/,
+      loader: ExtractTextPlugin.extract(
+        'style', 
+        'css?modules&importLoaders=1&context=' + paths.appSrc + '&localIdentName=[path]___[name]__[local]___[hash:base64:5]!postcss-loader'
```
elisechant: can exclude `-loader` from `postcss-loader`, it's implied

config/webpack.config.dev.js
```diff
@@ -213,6 +220,9 @@ module.exports = [{
     // makes the discovery automatic so you don't have to restart.
     // See https://github.com/facebookincubator/create-react-app/issues/186
     new WatchMissingNodeModulesPlugin(paths.appNodeModules),
+    new ExtractTextPlugin('[name].css', {
```
elisechant: kill this

config/webpack.config.prod.js
```diff
@@ -251,7 +254,10 @@ module.exports = [{
   },
   externals: /^[a-z\-0-9]+$/,
   module: {
-    loaders: loaders
+    loaders: loaders.concat({
+      test: /\.css$/,
+      loader: 'ignore-loader'
+    })
   },
   plugins: [
```
elisechant: need to include Extract in plugins too 
from CRA: 
`new ExtractTextPlugin('static/css/[name].[contenthash:8].css'),`

config/webpack.config.dev.js
elisechant: also you could load in a .babelrc instead

server/routes/render.js
```diff
@@ -72,12 +73,14 @@ const render = (request, response) => {
   // TODO test this behaviour
```
elisechant: todo here

src/shared/SellerProfile/ReviewHeader.js
```diff
@@ -2,6 +2,8 @@ import React from 'react';
 import classNames from 'classnames';
 import get from 'lodash/get';
 
+import profile from './profile.css'; // eslint-disable-line no-unused-vars
```
elisechant: looks like you don't need the eslint-disable

src/shared/SellerProfile/ReviewHeader.js
```diff
@@ -26,14 +28,14 @@ const ReviewHeader = (props) => {
   };
 
   return (
-    <section className="seller-profile seller-profile__review-header">
+    <section styleName="profile.review-header">
       <div className="row ">
         <div className="col-sm-8">
-          <h1 tabIndex="-1">{name}</h1>
+          <h1 tabIndex="-1" styleName="profile.heading">{name}</h1>
```
elisechant: shouldn't this be styleName={profile.heading} ? or that's the syntax from loader you're using?

elisechant:
Should be a matter of just adding `?modules&importLoaders=1&localIdentName=[path][name]--[local]` to the css-loader in dev and build. Shouldn't need to change CRA too much here to get what you need. Everything here https://github.com/css-modules/webpack-demo should _just work_ with CRA.

config/webpack.config.dev.js
nathf: The reason we extract on dev is we use this build in an integrated env with the other python apps (this includes SSR). JS only option causes SSR to crash as it requires the `window` object

src/shared/SellerProfile/ReviewHeader.js
nathf: String is the documented syntax, then eslint thinks profile is not being used, will give object a try

config/webpack.config.prod.js
nathf: Not for the server, hence the ignore loader

src/shared/SellerProfile/ReviewHeader.js
nathf: Has to be a string apparently.

nathf:
@elisechant updated where applicable per our conversation

config/webpack.config.prod.js
elisechant: gotcha

src/shared/SellerProfile/ReviewHeader.js
elisechant: just thinking in the future you might want to call this something like `sheet` instead of `profile` given that your style includes are not declarative

PR #151, opened 2017-01-12
Fix missing form fields

matt-sm:


nathf:
We will need to make corresponding changes to profile, but they can be made later

PR #154, opened 2017-01-15
Update badge map, remove invalid

nathf:


src/shared/SellerProfile/ReviewHeader.js
```diff
@@ -19,9 +19,8 @@ const ReviewHeader = (props) => {
   } = props;
 
   const badgeTitle = {
-    indigenous: 'Indigenous',
-    nfp_social_enterprise: 'Not-for-profit / social enterprise',
-    product: 'Product based business',
```
maxious: This is still available in https://github.com/AusDTO/dto-digitalmarketplace-frontend/blob/e62367be9d862601f6ac14ad20c2b53eb238e848/src/bundles/SellerRegistration/components/BusinessInfoForm/BusinessInfoForm.js#L144 ?

nathf: For now yea, I guess we could temporarily leave the badge. It will inevitably disappear though.

PR #160, opened 2017-01-18
Feature/search card

nathf:


src/shared/Badges/Badges.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import classNames from 'classnames';
```
matt-sm: thats an interesting package

PR #162, opened 2017-01-19
Display signed agreements

matt-sm:


src/shared/SellerProfile/PrivateInfo.js
```diff
@@ -36,6 +39,13 @@ const PrivateInfo = (props) => {
             return (<p key={i}><b>{question}</b> {answer} - {details}</p>)
         })}
         </Row>
+        <Row title="Master Agreement" show={!isEmpty(signed_agreements)}>
+        {signed_agreements.map((agreement, i) => {
+            console.log(agreement);
```
nathf: This needs to be removed

PR #163, opened 2017-01-20
Fix/document validation save

nathf:
* Updated document validation
* Persist application after single document upload


src/bundles/SellerRegistration/components/DocumentsForm/DocumentsForm.js
```diff
@@ -125,7 +126,15 @@ class DocumentsForm extends BaseForm {
                           valid={form.valid}
                           onCustomSubmit={onSubmit}
                           validators={{
-                              documents: (documents) => documents && Object.keys(documents).length === 3
+                            documents: (documents = {}) => {
+                                let keys = Object.keys(documents);
+
+                                if (keys.length !== 3) {
```
matt-sm: i uploaded all docs but validation still triggers?

PR #171, opened 2017-01-30
Init products

matt-sm:


src/bundles/SellerRegistration/components/ProductsForm/ProductsForm.test.js
```diff
@@ -0,0 +1,32 @@
...
+
+test('mapStateToProps', () => {
+    let state = generateFormValidilityState(true);
+    let props = mapStateToProps(state);
+    expect(props).toEqual({ form: {"valid": true}, "formErrors": undefined, "mode": "add", "model": "productsForm", "returnLink": undefined});
+
+    state = generateFormValidilityState(false);
+    props = mapStateToProps(state);
+    expect(props).toEqual({ form: { valid: false },"formErrors": undefined, "mode": "add", "model": "productsForm", "returnLink": undefined});
+});
```
nathf: These tests are redundant, since the selector (formProps) you're using already has a test

src/bundles/SellerRegistration/components/ProductsForm/ProductsForm.js
```diff
@@ -0,0 +1,167 @@
...
+            id="products"
+            component={SubmitForm}
+            valid={form.valid}
+            onCustomSubmit={onSubmit}
+          >
+            {csrf_token && (
+              <input type="hidden" name="csrf_token" id="csrf_token" value={csrf_token} />
+            )}
+
+            {productsForm.products && productsForm.products.map((product, key) => {
```
nathf: `productForm.products` shouldn't ever be falsy - since you've defined it in your reducer. A better defensive check may be to ensure its an array (which would still capture those falsy cases if they were to happen)?

src/bundles/SellerRegistration/components/ProductsForm/ProductsForm.js
```diff
@@ -0,0 +1,167 @@
...
+  static propTypes = {
+    action: React.PropTypes.string,
+    csrf_token: React.PropTypes.string,
+    form: React.PropTypes.object.isRequired
+  }
+
+  onAdd(e) {
+      e.preventDefault();
+      const {model, createProduct, productsForm} = this.props;
+      createProduct(model, productsForm);
```
nathf: Just pass the products straight to `createProduct`.

PR #173, opened 2017-01-31
Feature/search filters

nathf:


src/bundles/Search/components/Catalogue/Catalogue.js
```diff
@@ -76,19 +93,24 @@ export class Catalogue extends React.Component {
...
               <h4>No results found</h4>
             ) : (
               search.results.map((result, i) => (
                 <Card {...result} key={i} />
               ))
             )}
+
             <hr/>
+
             <Pagination
```
matt-sm: dont need pagination when no results found

PR #184, opened 2017-02-02
Updated no search results, with query reset

nathf:


src/bundles/Search/redux/modules/search.js
```diff
@@ -49,6 +50,19 @@ export const scrubState = (state) => {
   }, result);
 }
 
+export const falseValues = (target) => {
```
matt-sm: interesting

PR #185, opened 2017-02-02
Add validLink validator to link fields

nathf:


matt-sm:
would you still be able to post an invalid link to the server?

nathf:
You could, but it will have to being with `http`, I'm not aware of anyway to use javascript from there

PR #188, opened 2017-02-03
Make registration link customisable

maxious:


src/bundles/Header/AuthWidget.js
```diff
@@ -13,7 +13,7 @@ export const AuthWidget = (props) => {
 
   return (
     <ul className="inline-links--inverted">
-      <li><a href={props.registerUrl}>Register</a></li>
+      <li><a href={props.registerUrl}>{props.registerText}</a></li>
```
nathf: Can there be a default prop for this?

PR #196, opened 2017-02-07
Router v4 beta

nathf:


src/bundles/CaseStudy/components/DomainList/DomainList.js
```diff
@@ -319,10 +311,10 @@ const mapDispatchToProps = (dispatch) => {
     },
     onDeleteCaseStudy: (dispatchActions, id) => {
       dispatch(actions.omit('caseStudyForm.case_studies', id));
-      dispatch(dispatchActions.submitApplication());
+      //dispatchActions.submitApplication();
```
matt-sm: uncomment - should be called

PR #197, opened 2017-02-07
Mobile search accordian

matt-sm:
Put the mobile search filter in an accordion

### DIFF
--- a/src/shared/Card/Card.css
+++ b/src/shared/Card/Card.css
@@ -13,8 +13,10 @@
 
 .card h3 {
   margin: 0 0 0.75rem;
+}
+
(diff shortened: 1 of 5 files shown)


src/shared/Card/Card.css
```diff
@@ -13,6 +13,8 @@
 
 .card h3 {
   margin: 0 0 0.75rem;
+  border-bottom: solid 1px #5bcbe3;
+  display:table;
```
nathf: table?

matt-sm: Easiest way to get the width of the h3 border same as the text

PR #198, opened 2017-02-07
Ensure terms are accepted when submitting application

maxious:


src/bundles/SellerRegistration/redux/modules/submitStep.js
```diff
@@ -0,0 +1,9 @@
+const initialState = {};
```
nathf: This whole file is redundant?

src/bundles/SellerRegistration/components/Signup/Signup.js
```diff
@@ -133,7 +134,7 @@ class Signup extends React.Component {
 
   render() {
     const { forms, router, steps = {}, location } = this.props;
-    const applicationValid = (this.steps.length - 1) === Object.keys(steps).length;
+    const applicationValid = (this.steps.length - 1) < Object.keys(steps).length;
```
nathf: Why less than?

PR #208, opened 2017-02-08
Feature/skip case studies

nathf:


src/bundles/SellerRegistration/components/DomainSelector/DomainSelector.js
```diff
@@ -40,6 +45,16 @@ class DomainSelector extends BaseForm {
             <Layout>
                 {header}
                 <article role="main">
+                    <Link
+                        to="/products"
+                        onClick={() => {
```
maxious: Could we reevaluate casestudy step when services/digital step changes?

PR #216, opened 2017-02-12
Date Field validation

maxious:


matt-sm:
- when i have uploaded all docs, but there is a validation error on a date, I still see 'please upload all documents'

src/validators.js
```diff
@@ -12,6 +15,18 @@ export const required = (val) => {
   return val && val.trim().length;
 };
 
+export const validDate = (val) => {
```
matt-sm: unit test for this validator?

maxious: added some unit tests

PR #220, opened 2017-02-13
remove found on result count, hide sort by, add casestudies tab

djrobstep:
superficial interface changes as requested by N


matt-sm:
Is the intention to show sort by at some point?

djrobstep:
N wanted it gone temporarily, but we presumably will want to show it again at some point

PR #222, opened 2017-02-13
Recruiter form init

matt-sm:
- delete recruitment domain
- add new recruiter step


matt-sm:
- check box, then uncheck
- save and continue
- no errors/validation errors but cannot proceed

PR #245, opened 2017-02-21
Fix case study and document views on profile

maxious:


src/shared/SellerProfile/Body.js
```diff
@@ -266,17 +265,11 @@ const Body = (props) => {
...
                 <tr key={val}>
                   <td>
-                    {public_profile ?
-                      documentTitle[key]
-                    : (
-                      <a href={`${documentsUrl}${filename}`}>
-                      {documentTitle[key]}
-                      </a>
-                    )}
+                    {documentTitle[key]}
```
matt-sm: now when you preview your profile you cant check your docs

maxious: Correct. That was the intent.

PR #250, opened 2017-02-22
enable seller registration declaration and remove placeholder

maxious:


src/bundles/SellerRegistration/components/Finish.js
```diff
@@ -1,24 +0,0 @@
-import React from 'react';
```
matt-sm: i wouldn't delete this page just yet - we may have to re-enable if things go wrong

PR #266, opened 2017-02-24
Validate recruitment fields as numerical percentages

maxious:


src/__tests__/validators.test.js
```diff
@@ -53,6 +53,21 @@ test('validPhoneNumber', () => {
     expect(validator.validPhoneNumber('+61 2 12345678')).toBeTruthy();
 })
 
+test('validPercentage', () => {
+
+    expect(validator.validPercentage(false)).toBeTruthy();
+
+    expect(validator.validPercentage('1 ')).toBeFalsy();
+    expect(validator.validPercentage('1% ')).toBeFalsy();
```
matt-sm: '1% ' should be true

maxious: allowing space now.

PR #267, opened 2017-02-24
link updates including profile preview insurance documents

maxious:


matt-sm:
Docs can now come out of PrivateInfo

maxious:
Finance is still only linked there though so might as well leave.

PR #272, opened 2017-03-01
Try to fix address errors

matt-sm:


src/bundles/SellerRegistration/redux/modules/reducer-signup.js
```diff
@@ -37,7 +37,14 @@ export default combineReducers({
       summary: '',
       website: '',
       linkedin: '',
-      addresses: {}
+      addresses: {
```
maxious: Is this an array or an object?

matt-sm: object with the {"0":{}, "1":{}} format

PR #274, opened 2017-03-06
readonly abn field for existing sellers

maxious:


src/bundles/SellerRegistration/components/BusinessDetailsForm/BusinessDetailsForm.js
```diff
@@ -89,7 +88,7 @@ class BusinessDetailsForm extends BaseForm {
                               (<span>You need an Australian Business Number to do business in Australia.&nbsp;
                               <a href='https://abr.gov.au/For-Business,-Super-funds---Charities/Applying-for-an-ABN/Apply-for-an-ABN/' target="_blank" rel="external">Apply for an ABN here.</a>
                           </span>)}
-                          disabled="disabled"
+                          readOnly={!!supplierCode}
```
matt-sm: is !! an actual operator?

src/shared/form/Textfield.js
```diff
@@ -1,55 +1,63 @@
...
-      pattern={pattern}
-      maxLength={maxLength}
-      disabled={disabled}
-    />
-  </div>
-);
+class Textfield extends React.Component {
+    render() {
+        const {name, id, htmlFor, label, model, validators, messages, description, pattern, maxLength, disabled, readOnly, type} = this.props;
+        const readonly = (readOnly ? {"readOnly": true, "disabled": true} : {});
```
matt-sm: why not make the props we pass to our Textfield match what Control supports?

src/bundles/SellerRegistration/components/BusinessDetailsForm/BusinessDetailsForm.js
maxious: yes, webpack told me to use it - !  converts the object being not null/null to true/false and then another negates that ie. readOnly is true if supplierCode is not null

matt-sm: cool - didn't know that!

src/shared/form/Textfield.js
maxious: I've split disabled and readOnly so they match all the way through

PR #276, opened 2017-03-06
Make government experience question mandatory

maxious:


src/bundles/SellerRegistration/components/BusinessInfoForm/BusinessInfoForm.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import {connect} from 'react-redux';
-import {Form, Control} from 'react-redux-form';
+import {Form, Fieldset, Control, Errors} from 'react-redux-form';
```
matt-sm: Errors not used

PR #292, opened 2017-03-20
compare step keys to determine application completion

djrobstep:


src/bundles/SellerRegistration/components/Signup/Signup.js
```diff
@@ -148,7 +149,18 @@ class Signup extends React.Component {
...
-    const applicationValid = (this.filteredSteps.length - 1) <= values(steps).filter(s => s === "complete").length;
+
+    let stepKeys = this.filteredSteps.map(s => s['id']);
+    stepKeys = stepKeys.filter(s => s !== 'start' && s != 'submit');
+    let doneKeys = keys(steps).filter(s => steps[s] === "complete" && s !== 'start' && s != 'submit');
+
+    let stepsSet = new Set(stepKeys);
+    let doneSet = new Set(doneKeys);
+
+    let difference = new Set([...stepsSet].filter(x => !doneSet.has(x)));
```
maxious: Could this difference object be used to display a warning on the submit step?

djrobstep: Yep that's the next step :)

PR #297, opened 2017-03-22
Fix crossplatform line breaks in long text display

maxious:


src/helpers.js
```diff
@@ -0,0 +1,7 @@
+export const newline = (val) => {
```
matt-sm: nitpicker - the function name doesn't describe what its doing

PR #298, opened 2017-03-22
Fix supplier code and add some tests

matt-sm:
- Had to use isNumber() to cater for supplierCode 0
- Minor tweaks to some components that didn't previously have unit tests

### DIFF
--- a/src/bundles/SellerRegistration/components/BusinessDetailsForm/BusinessDetailsForm.js
+++ b/src/bundles/SellerRegistration/components/BusinessDetailsForm/BusinessDetailsForm.js
@@ -2,6 +2,7 @@ import React from 'react';
 import {connect} from 'react-redux';
 import { Form, actions } from 'react-redux-form';
 import isEmpty from 'lodash/isEmpty';
+import isNumber from 'lodash/isNumber';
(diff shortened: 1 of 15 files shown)


src/bundles/SellerRegistration/components/BusinessDetailsForm/BusinessDetailsForm.js
```diff
@@ -85,7 +86,7 @@ class BusinessDetailsForm extends BaseForm {
                           id="abn"
                           htmlFor="abn"
                           label="ABN"
-                          description={supplierCode ? "You need an Australian Business Number to do business in Australia." :
+                          description={isNumber(supplierCode) ? "You need an Australian Business Number to do business in Australia." :
                               (<span>You need an Australian Business Number to do business in Australia.&nbsp;
                               <a href='https://abr.gov.au/For-Business,-Super-funds---Charities/Applying-for-an-ABN/Apply-for-an-ABN/' target="_blank" rel="external">Apply for an ABN here.</a>
                           </span>)}
```
maxious: do readonly and disabled need to be updated too?

PR #300, opened 2017-03-27
buyer teamview

djrobstep:


src/bundles/BuyerDashboard/components/Overview/Overview.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import { connect } from 'react-redux'
+
+import './Overview.css'
+
+const Overview = ({meta = {}}) => (
```
matt-sm: Do you need to pass that prop?

src/bundles/BuyerDashboard/components/Overview/Overview.js
```diff
@@ -0,0 +1,23 @@
...
+)
+
+const mapStateToProps = ({meta}, ownProps) => {
+  return {
+    ...ownProps,
+    meta
+  };
+};
+
+const mapDispatchToProps = (dispatch) => {
```
matt-sm: delete?

PR #317, opened 2017-04-06
Seller rego noproducts nav

Blackfry:
Navigation changes depending on products/services selected. Tests

### DIFF
--- a/src/bundles/SellerRegistration/ApplicantSignup.json
+++ b/src/bundles/SellerRegistration/ApplicantSignup.json
@@ -35,8 +35,8 @@
       }
     },
     "services": {
-        "Change & transformation": true,
-        "Cyber security": true
+        "Change & transformation": false,
+        "Cyber security": false
     },
     "case_studies": {
         "f29a3cbd-5d76-4fb1-f4ea-40e1b196a84e": {
--- a/src/bundles/SellerRegistration/components/ProductsForm/ProductsForm.js
+++ b/src/bundles/SellerRegistration/components/ProductsForm/ProductsForm.js
@@ -48,8 +48,8 @@ class ProductsForm extends BaseForm {
   }
 
   onClearProducts() {
-    const {clearProducts} = this.props;
-    clearProducts(this.props)
+    const {clearProducts, services} = this.props;
+    clearProducts(services)
   }
 
   render() {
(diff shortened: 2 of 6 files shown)


src/bundles/SellerRegistration/ApplicantSignup.json
```diff
@@ -35,8 +35,8 @@
       }
     },
     "services": {
-        "Change & transformation": true,
-        "Cyber security": true
+        "Change & transformation": false,
```
matt-sm: this is needed for a test?

Blackfry: I assumed that is dummy data and not relevant to anything else. I had to set them to false to check navigation from products when there are no services.

src/bundles/SellerRegistration/components/ProductsForm/ProductsForm.js
```diff
@@ -43,6 +47,14 @@ class ProductsForm extends BaseForm {
       removeProduct(model, id);
   }
 
+  onEmptyNav() {
+    const {navFromProducts} = this.props;
+    let servicesEmpty = !some(Object.values(this.props.services));
```
matt-sm: some() beacuse of the case {servicex: false} ?

Blackfry: Yes. Services that have been selected, then unselected, get flipped to false, not removed so need to check for trues.

src/bundles/SellerRegistration/redux/modules/steps.js
```diff
@@ -18,20 +19,32 @@ export default function reducer(state = {}, action = {}) {
       return Object.assign({}, state, {
         [step]: STATUS.partial
       });
+    case STEP_CLEAR:
+      delete state[step]
+      return Object.assign({}, state, {
```
matt-sm: rather than deleting could use void 0:
`Object.assign({}, state, {[step]: void 0}`

src/bundles/SellerRegistration/redux/modules/steps.js
```diff
@@ -18,20 +19,32 @@ export default function reducer(state = {}, action = {}) {
       return Object.assign({}, state, {
         [step]: STATUS.partial
       });
+    case STEP_CLEAR:
+      return Object.assign({}, state, {
+        ['digital']: void 0
```
matt-sm: should pass this as a parameter - dont hardcode

PR #320, opened 2017-04-10
Fix documents section seller profile

Blackfry:
changes to seller profile


src/shared/SellerProfile/Body.js
```diff
@@ -65,14 +65,14 @@ const Body = (props) => {
                     </div>
                 )}
 
-
+                /* Hide learn about assessments link MAR-490
```
matt-sm: perhaps delete not comment

PR #322, opened 2017-04-11
Hide learn about assessments ink

Blackfry:
Made this link optional based on state of assessed/unassessed

### DIFF
--- a/src/shared/SellerProfile/Body.js
+++ b/src/shared/SellerProfile/Body.js
@@ -65,6 +65,15 @@ const Body = (props) => {
                     </div>
                 )}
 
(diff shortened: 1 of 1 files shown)


src/shared/SellerProfile/Body.js
```diff
@@ -65,6 +65,15 @@ const Body = (props) => {
                     </div>
                 )}
 
+                {isEmpty(assessed || !isEmpty(unassessed)) && (
```
matt-sm: is that a typo?

Blackfry: yep

PR #335, opened 2017-04-21
changes for seller catalogue search

Blackfry:


src/shared/Card/Card.js
```diff
@@ -6,61 +6,64 @@ import Badges from '../Badges';
...
           <p styleName="description">{description}</p>
-            {/* TODO refactor when products exists
-             <div className="products">
-             <Icon value="product" size={18} /> <strong>Digital products:</strong> CMS
-             </div>
-             */}
-          {view === "products" && <div styleName="product-links">
-               <span>
+          {/* TODO refactor when products exists
+           <div className="products">
```
maxious: Products does exist, don't need this commented out code :)

PR #360, opened 2017-05-09
Add divider for link title field

maxious:


matt-sm:
There seems to be more going on in here than just adding a divider.

PR #367, opened 2017-05-14
Manage team on seller dashboard

Blackfry:
styling and fix dates on team brief view


src/bundles/BuyerDashboard/components/Dashboard/BuyerDashboard.css
```diff
@@ -37,6 +32,7 @@
 
 .filters {
   margin: 1.5rem 0 0;
+  width: 74.48861%
```
matt-sm: thats a very precise percentage

Blackfry: That's precision! It was a number that was being used and I couldn't see that it was calculated anywhere, it worked though.

src/bundles/BuyerDashboard/components/DashboardBriefs/DashboardBriefs.css
```diff
@@ -1,60 +1,184 @@
...
     }
 
-    .summary-item-field-heading-first {
-        vertical-align: bottom !important;
+    table.brief-table thead tr th.right-header {
+        width: 10%
     }
 }
 
+table.brief-table {
```
matt-sm: is there some duplication in these table styles?

Blackfry: It replaces styling from "content-tables" from a uikit style sheet. I don't think there's any doubling up within this file though.

src/bundles/BuyerDashboard/components/Dashboard/BuyerDashboard.css
maxious: these exact percentages are rems or ems or something.

PR #369, opened 2017-05-16
Edit profile assessments admin

Blackfry:
Highlight if changes were made to seller profile case studies


src/bundles/SellerRegistration/components/ApplicationPreview/ApplicationPreview.js
```diff
@@ -23,6 +54,9 @@ const ApplicationPreview = ({ header, body, privateInfo, onClick }) => (
         <Link to="/review">Go back and edit</Link>
       </div>
     </div>}
+    <div>
+      <Changes {...body}/>
```
maxious: Does this show up in the public view of the profile? You could use the same logic as PrivateInfo to hide {!onClick && privateInfo.disclosures

Blackfry: it renders conditionally on unassessed case studies which public profile shouldn't have but I'll take a look at that

matt-sm:
there are more changes after discussion with Neil?

Blackfry:
This was going to v1 with v2 pulling out details of changes made via json diffing. Depends if they are happy to wait for that

PR #380, opened 2017-05-24
preview of full page rendering on node

matt-sm:
- this successfully renders localhost:60000/collaborate
- piggy-backing on what we already have with server component rendering
- work's side-by-side with existing python rendering

todos before it could go live:
- header/footer
- the rest of everything from the flask templates (scripts, etc)
- stylesheet? it currently uses main.css from this repo
- 500/404 handling in express


server/routes/render.js
```diff
@@ -73,26 +65,77 @@ const render = (request, response) => {
...
+        component = renderComponent(route.widgetPath, props, false);
+      }
+      return match
+    })
+
+    if (!validRoute) {
+      return response.status(404).send('Page not found');
+    }
+
+    response.send('<!doctype html>' + ReactDOMServer.renderToString(
```
maxious: I did not know this was a thing... https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/HTML5/Introduction_to_HTML5#Declaring_that_the_document_contains_HTML5_mark-up_with_the_HTML5_doctype

maxious: If you ever want a speedup... https://github.com/aickin/react-dom-stream

matt-sm: react-dom-stream does look interesting

PR #381, opened 2017-05-25
Swap out rrf form with plain react component. Add email template text.

Blackfry:
Change some copy


src/bundles/ApplicationsAdmin/components/AppList/AppList.js
```diff
@@ -95,7 +114,7 @@ class AppList extends Component {
                     onAcceptClick(a.id);
                   }} name="Accept">Accept</button>
                   }
-                  { (a.status === 'submitted' && a.type === 'edit') &&
+                  { a.status === 'submitted' &&
```
matt-sm: wouldn't you want to leave this as-is?

src/bundles/ApplicationsAdmin/components/RevertNotification/RevertNotification.js
```diff
@@ -1,92 +1,114 @@
...
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  handleChange(e) {
+    this.setState({
+      message: e.target.value
+    })
+  }
+
+  resetMessageText(propsMessage) {
```
matt-sm: could you combine this in onClose()?  it will always reset to templateString?

src/bundles/ApplicationsAdmin/components/RevertNotification/RevertNotification.js
```diff
@@ -1,92 +1,114 @@
...
-      </Layout>
+      <form>
+        <label htmlFor="formHeading">
+          <h3>
+            Revert application with optional email to seller
+          </h3>
+          <b>
+            {`Provide a short clear reason, starting with a hyphen. Eg "${"- ABN missing a digit"}"`}
+          </b>
+          <textarea value={this.state.message} onChange={this.handleChange}/>
```
matt-sm: doesn't seem quite right the way this is setup - why put textare value in state rather than submit form? Hang on are you following https://facebook.github.io/react/docs/forms.html?

PR #388, opened 2017-05-30
App changes field validation

Blackfry:
Doubles the number of ignored fields re application changes list


src/bundles/ApplicationsAdmin/components/AppChanges/helper.js
```diff
@@ -1,4 +1,5 @@
 var jsonpatch = require('fast-json-patch');
+import isEmpty from 'lodash/isEmpty';
 
 // filter out application changes to keys we don't care about
 let ignoreKeys = ["/supplier_code", "/supplier", "/submitted_at", "/status", "/seller_types/recruitment",
```
matt-sm: I am curious why keys like `/case_studies/11e27ef8-691b-38c9-4691-f48e6eef031e, /case_studies/4/links/supplier`

PR #418, opened 2017-06-21
Seller scenario messaging

Blackfry:
added assessment progress ui

### DIFF
--- a/src/bundles/CaseStudy/components/SubmitConfirmation/SubmitConfirmation.js
+++ b/src/bundles/CaseStudy/components/SubmitConfirmation/SubmitConfirmation.js
@@ -3,28 +3,111 @@ import {connect} from 'react-redux';
 import format from 'date-fns/format';
 import startCase from 'lodash/startCase';
 import Icon from '../../../../shared/Icon';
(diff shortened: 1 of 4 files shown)


src/bundles/CaseStudy/components/SubmitConfirmation/SubmitConfirmation.js
```diff
@@ -12,28 +13,116 @@ class SubmitConfirmation extends React.Component {
...
+            profileUrl,
+            unassessedDomains,
+            assessedDomains
         } = this.props;
 
+        let hasUnassessed = !isEmpty(unassessedDomains) && Object
+            .keys(unassessedDomains)
+            .length > 0;
+        
+        let hasAssessed = !isEmpty(assessedDomains) && assessedDomains;
```
matt-sm: is the `&& assessedDomains` required?

Blackfry: This component is being targeted by two flows so assessedDomains may not be passed in, so I'm checking in one case where it is and it's empty and the other where we don't expect it

PR #427, opened 2017-07-06
Add sections to edit details

marcoalfonso:


src/shared/form/Textfield.js
```diff
@@ -5,14 +5,16 @@ import StatefulError from './StatefulError';
 
 class Textfield extends React.Component {
     render() {
-        const {name, id, htmlFor, label, model, validators, messages, description, pattern, maxLength, disabled, readOnly, type} = this.props;
+        const {name, id, htmlFor, label, model, validators, messages, description, pattern, maxLength, disabled, readOnly, type, descriptionWithHtml} = this.props;
```
marcoalfonso: Isn't that a bit of an overkill? The current description then could also be a component...

matt-sm: Seems cleaner with a component but maybe overkill

PR #432, opened 2017-07-14
Mar 731 refactor signup page

marcoalfonso:


app/components/Header/Header.css
```diff
@@ -0,0 +1,3 @@
+.join {
```
matt-sm: why does this component have .css where others have .scss?

marcoalfonso: Ah removed, was waiting for the header work to start using scss on header.

matt-sm:
General comment - should we do some unit tests?

marcoalfonso:
Yeah unit tests for sure, but maybe once the whole page is completed

PR #433, opened 2017-07-17
Refactor signup page with api call and success message

marcoalfonso:


src/bundles/SellerRegistration/components/SignupForm/SignupForm.js
```diff
@@ -20,121 +22,161 @@ class SignupForm extends BaseForm {
...
+    if (response.status >= 200 && response.status < 300) {
+      return response
+    } else {
+      var error = new Error(response.statusText)
+      error.response = response
+      throw error
+    }
+  }
+
+  handleSubmit(model) {
```
matt-sm: why not use redux to do this?

PR #435, opened 2017-07-18
More responsive search

maxious:


src/bundles/Search/redux/modules/search.js
```diff
@@ -194,7 +195,7 @@ export const search = (type, value, options = {}) => {
       .then(res => res.json())
       .then(json => dispatch(syncResult(json)))
       .catch(() => dispatch({ type: ERROR_SEARCH }));
-    }, 500);
+    }, 250);
```
matt-sm: what is this magic number?

maxious: this is a debouncer, the magic number is how many ms to wait for the input to stop coming before starting the search API call

PR #437, opened 2017-07-19
Use "1" to mean "1" when counting employees

sarneaud:
"Sole trader" has a special meaning in Australian business law:
https://www.business.gov.au/info/plan-and-start/start-your-business/business-structure/business-structures-and-types/sole-trader

Not all single-person business entities are sole traders.


matt-sm:
We'll need to ask business/legal people about this one

PR #454, opened 2017-08-02
React router

Blackfry:
Routing SignupForm with rr4


app/components/shared/NotFound.js
```diff
@@ -0,0 +1,21 @@
...
+
+const NotFound = () => (
+  <div className="error-page">
+    <div className="index-page grid-row">
+      <div className="column-two-thirds">
+        <header className="page-heading-smaller">
+          <h1>Page not found</h1>
+        </header>
+        <p>
+          Check you've entered the correct web address or start again on the <a href="/">Digital Marketplace homepage</a>.
```
marcoalfonso: Can you escape the quote so we don't get the editor complaining? 🙏

app/store/index.js
```diff
@@ -1,11 +1,17 @@
...
 
 export default function configureStore(initialState) {
-  return createStore(
-    rootReducer,
-    initialState,
-    applyMiddleware(thunk)
-  );
-}
\ No newline at end of file
+  const debugware = [];
```
marcoalfonso: Do we really need this here? if you want it for dev debugging then create a production store that doesn't have it.

PR #461, opened 2017-08-07
Update dependencies for Node8

maxious:


package.json
```diff
@@ -196,6 +191,9 @@
       "modules": false,
       "location": "pancake/js/",
       "name": "pancake.min.js"
+    },
+    "react": {
```
marcoalfonso: What is this for?

maxious: pancake-react!

src/bundles/SellerRegistration/components/SignupForm/SignupForm.js
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
```
matt-sm: we should remove (and a few other files) this as its now in `app/`

maxious: need to remove from buyer app and buyer app tests first.

PR #463, opened 2017-08-08
Fix/signup accessibility and hotfixes

marcoalfonso:


app/components/SignupForm/SignupForm.js
```diff
@@ -147,14 +148,27 @@ class SignupForm extends BaseForm {
...
+  }
+
   render() {
     const { csrf_token, model, form, children, signupForm, buyer_url, seller_url } = this.props
     let valid = form.valid
     let employmentStatus = signupForm.employment_status
     let action = isBuyer ? buyer_url : seller_url
     let { signupSuccess, signupMessage, isBuyer, emailValidators, emailErrorMessages } = this.state
 
-    /*eslint-disable indent */
```
maxious: 👍

app/components/SignupForm/SignupForm.js
```diff
@@ -37,10 +37,11 @@ class SignupForm extends BaseForm {
       },
       emailErrorMessages: {
         required: 'Your email is required',
-        validEmail: 'A validly formatted email is required',
-        governmentEmail: 'Email should have a government domain'
+        validEmail: 'A validly formatted email is required.',
```
matt-sm: Neil says:
> You must provide a valid email address

marcoalfonso: Huh? Where does it say that? 😕

PR #467, opened 2017-08-09
CSS modules

marcoalfonso:


app/components/Banner/Banner.scss
```diff
@@ -1,14 +1,16 @@
-.phase-banner {
+@import "../../App.scss";
```
matt-sm: we need to do this at the top of every component scss?

app/components/SignupForm/SignupForm.scss
```diff
@@ -1,53 +0,0 @@
-.user-type .radio-list-container{
```
matt-sm: 👍

app/components/Banner/Banner.scss
marcoalfonso: No, its just if you want to include global app styles like in this case wrapper. Ultimately everything should be self contained except for layouts, etc.

PR #469, opened 2017-08-11
Fix accessibility

marcoalfonso:


app/components/SignupForm/SignupForm.js
```diff
@@ -279,6 +279,7 @@ class SignupForm extends BaseForm {
                       id="name"
                       htmlFor="name"
                       label="Full name"
+                      ariaDescribedBy="Full Name"
```
maxious: doesn't aria-describedby point to the ID of a <label>? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute

marcoalfonso: Good catch, let me fix 👍

app/components/shared/form/Textfield.js
```diff
@@ -19,12 +19,13 @@ class Textfield extends React.Component {
       maxLength,
       disabled,
       readOnly,
-      type
+      type,
+      ariaDescribedBy
```
maxious: now you don't use this prop.

PR #470, opened 2017-08-11
New create user

Blackfry:


app/actions/memberActions.js
```diff
@@ -16,28 +20,148 @@ export function memberInfoIsLoading(bool) {
...
+        data: response.data
+      }
+
+    case 409:
+      return {
+        type: CREATE_USER_DUPLICATE_FAILURE,
+        errorMessage: response.data.message
+      }
+
+    case 400:
```
matt-sm: there could be other flavours of 4xx returned as well - 400 and 500 are redundant anyway if default is an error

Blackfry: The auth/view create user explicitly returns 200, 400, 409 and 500 but yes there is little value in handling 400 and 500 instead of having them fall through to default

package.json
```diff
@@ -53,6 +53,7 @@
     "jest": "^20.0.4",
```
matt-sm: shouldn't `redux-logger` be in `package.json`?

PR #483, opened 2017-08-21
show message if js disabled

matt-sm:


views/index.ejs
```diff
@@ -47,6 +47,9 @@
             });
           ga('require', 'urlChangeTracker');
     </script>
+    <noscript>
+     Javascript is disabled. Please enable javascript and refresh the page.
```
elisechant: You probably could have styling this or at least pop it in a `<p>`. As well, think the messaging could be improved, perhaps include a link to "how to enable JavaScript" something like http://www.enable-javascript.com/ perhaps?

matt-sm: Totally agree @elisechant .  We will need to style this and provide proper instructions.  I just wanted to do a quick experiment for a user who I believe has js disabled.

PR #486, opened 2017-08-22
use /2 as new react root path

Blackfry:
move createUser out of /signup


matt-sm:
there would be a corresponding api + buyer PR?

PR #492, opened 2017-08-24
Error messages

matt-sm:
- ~common component for errors from actions~
- uses existing `ErrorBox` to display api errors
- decreases amount of error state

TODO
- needs to be moved out of user actions
- ~focus/scroll when error displayed~
- ~when to "clear" errors?~


app/components/CreateUser/CreateUser.js
```diff
@@ -11,16 +12,15 @@ export class CreateUser extends Component {
...
 
     return (
       <div className="row">
         <div className="col-sm-push-2 col-sm-8 col-xs-12">
-          {loadSignupSuccess &&
-            <div>
-              <CreateUserForm initialState={userRegisterDetails} />
-            </div>}
-          {loadSignupErrored && <div>error loading user registration details</div>}
+          <article role="main">
```
marcoalfonso: Why are you using article? Coding html like it is 1999 :P?

matt-sm: I'm just copying what was already there :)

Should it be main?

app/actions/memberActions.js
```diff
@@ -155,11 +113,22 @@ export function createUser(values) {
...
       }
       dispatch(handleDataLoading(false))
     })
   }
 }
+
+export function setErrorMessage(message) {
+  return {
+    type: SET_ERROR_MESSAGE,
+    message: message
```
Blackfry: can be simplified to `message`

app/components/SignupForm/SignupForm.js
```diff
@@ -88,29 +86,13 @@ class SignupForm extends BaseForm {
...
-          : <li>
-              <p>
-                The Digital Marketplace encountered an error trying to send your signup email. Please try again later or{' '}
-                <a href="/contact-us" target="_blank" rel="external">
-                  {' '}contact us{' '}
-                </a>{' '}
-                for assistance.
-              </p>
-            </li>
+        signupSuccess: nextProps.signupSuccess
```
Blackfry: with the html removed from `signupMessage` we should be able to remove signupSuccess from component state and rely on `this.props.signupSuccess`

PR #495, opened 2017-08-25
Add rollbar source map upload and babel-polyfill for IE11

maxious:


matt-sm:
How much does the prod bundle get increased by with the rollbar source maps?

PR #497, opened 2017-08-29
Feature/spinner

marcoalfonso:


app/components/LoadingButton/LoadingButton.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * LoadingButton.js
+ *
+ * Wraps the loading indicator in a tag with the .btn--loading class
+ */
+
+import React from 'react'
+import LoadingIndicator from '../../components/LoadingIndicator/LoadingIndicator'
+
+function LoadingButton(props) {
```
Blackfry: are we doing es6 per FE meeting?

matt-sm: We should wait until we finalise the Airbnb style rules we will follow

app/reducers/appReducer.js
```diff
@@ -0,0 +1,42 @@
...
+    case SENDING_REQUEST:
+      return {
+        ...state,
+        currentlySending: action.sending
+      }
+
+    case SET_ERROR_MESSAGE:
+      return {
+        ...state,
+        message: action.message
```
matt-sm: You need to change `ErrorBox` cause it's still reading from user.message

Should we also change this to `action.errorMessage`?

matt-sm:
Maybe beyond scope of this, but what about Forms that need to call api to pre-load data?

PR #501, opened 2017-08-31
Reset user password

Blackfry:
duplicate commits here from when I rebased


app/components/ResetPassword/RequestResetEmailForm.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react'
+import { Form } from 'react-redux-form'
+import ErrorBox from '../../components/shared/form/ErrorBox'
+import Textfield from '../../components/shared/form/Textfield'
+import PageAlert from '@gov.au/page-alerts'
+import { required, validEmail } from '../validators'
+
+export const RequestResetEmailForm = ({ model, form, user, submitClicked, handleSubmit }) => {
```
matt-sm: the pattern I propose we follow:
```
const RequestRestEmailForm = props => {
  const { model, form, user, submitClicked, handleSubmit } = props
  return (
    // JSX
  )
}

RequestResetEmailForm.propTypes = {
  // type for each prop
}

export default RequestResetEmailForm
```

app/pages/ResetPasswordPage.js
```diff
@@ -0,0 +1,159 @@
...
+        form={form}
+        user={user}
+        submitClicked={this.onSubmitClicked}
+        handleSubmit={handleSubmit}
+      />
+    )
+  }
+}
+
+RequestResetEmail.propTypes = {
```
matt-sm: missing props

app/pages/ResetPasswordPage.js
```diff
@@ -0,0 +1,159 @@
...
+import { connect } from 'react-redux'
+import BaseForm from '../components/shared/form/BaseForm'
+import formProps from '../components/shared/form/formPropsSelector'
+import { RequestResetEmailForm } from '../components/ResetPassword/RequestResetEmailForm'
+import { ResetPasswordForm } from '../components/ResetPassword/ResetPasswordForm'
+import { sendResetPasswordEmail, submitResetPassword, getUserDataFromToken } from '../actions/resetPasswordActions'
+import { setErrorMessage } from '../actions/appActions'
+
+class RequestResetEmail extends BaseForm {
+  static propTypes = {
```
matt-sm: kill these static prototypes - you've got them down below

app/pages/ResetPasswordPage.js
```diff
@@ -0,0 +1,159 @@
...
+
+const mapDispatchToProps = dispatch => {
+  return {
+    handleSubmit: model => dispatch(sendResetPasswordEmail(model))
+  }
+}
+
+export const RequestResetEmailContainer = withRouter(connect(mapStateToProps, mapDispatchToProps)(RequestResetEmail))
+
+class ResetPassword extends BaseForm {
```
matt-sm: separate file?

app/pages/ResetPasswordPage.js
```diff
@@ -0,0 +1,159 @@
...
+
+const mapResetDispatchToProps = dispatch => {
+  return {
+    handleSubmit: payload => dispatch(submitResetPassword(payload)),
+    loadInitialData: token => dispatch(getUserDataFromToken(token)),
+    passwordsMatchMessage: message => dispatch(setErrorMessage(message))
+  }
+}
+
+export const ResetPasswordContainer = withRouter(connect(mapResetStateToProps, mapResetDispatchToProps)(ResetPassword))
```
matt-sm: I don't get where this is supposed to fit in - shouldn't it be part of the page?

app/routes.js
```diff
@@ -14,12 +15,13 @@ const RootContainer = () =>
...
 
-export const Routes = ({ match }) =>
+export const Routes = () =>
   <Switch>
-    <Route exact path={match.url} component={SignupContainer} />
+    <Route exact path={rootPath} component={SignupContainer} />
     <Route path={`${rootPath}/signup`} component={SignupContainer} />
     <Route path={`${rootPath}/createuser/:tokenstring`} component={CreateUserPage} />
     <Route path={`${rootPath}/brief/:brief_id/respond`} component={BriefResponse} />
+    <Route path={`${rootPath}/reset-password`} component={ResetPasswordPage} />
```
matt-sm: we've got dashes in this route but not createuser?

app/pages/ResetPasswordPage.js
```diff
@@ -0,0 +1,119 @@
...
+
+          <Route
+            path={`${match.url}/:tokenString`}
+            render={() =>
+              <ResetPasswordForm
+                model={model}
+                form={form}
+                user={user}
+                submitClicked={this.onSubmitClicked}
+                handleSubmit={values => this.handleResetPasswordSubmit(values)}
```
matt-sm: why is this function on the class where the other `handleSubmit` was a prop?

app/pages/ResetPasswordPage.js
```diff
@@ -0,0 +1,119 @@
+import React, { Component } from 'react'
```
matt-sm: I like it 👍

app/routes.js
```diff
@@ -3,24 +3,19 @@ import { withRouter, Switch, Route } from 'react-router-dom'
...
-    <Route path={rootPath} component={Routes} />
-    <Route exact path="/signup/createuser/:tokenstring" component={CreateUserPage} />
-    <Route component={NotFound} />
-  </Switch>
-
-export default withRouter(RootContainer)
+export const rootPath = '/2'
 
-export const Routes = ({ match }) =>
+export const RootContainer = () =>
```
matt-sm: 🎉

app/routes.js
```diff
@@ -3,24 +3,19 @@ import { withRouter, Switch, Route } from 'react-router-dom'
...
   <Switch>
-    <Route exact path={match.url} component={SignupContainer} />
+    <Route exact path={rootPath} component={SignupContainer} />
     <Route path={`${rootPath}/signup`} component={SignupContainer} />
-    <Route path={`${rootPath}/createuser/:tokenstring`} component={CreateUserPage} />
+    <Route path={`${rootPath}/createuser`} component={CreateUserPage} />
     <Route path={`${rootPath}/brief/:brief_id/respond`} component={BriefResponse} />
+    <Route path={`${rootPath}/reset-password`} component={ResetPasswordPage} />
```
matt-sm: still inconsistent with our routes though

app/pages/ResetPasswordPage.js
Blackfry: the function computes form values while the prop is a straight dispatch action

app/actions/resetPasswordActions.js
```diff
@@ -0,0 +1,94 @@
+import {
```
matt-sm: I suggest we start the fat arrow migration with these files

app/actions/resetPasswordActions.js
```diff
@@ -0,0 +1,94 @@
...
+  return dispatch => {
+    dispatch(sendingRequest(true))
+    dmapi({ url: `/reset-password/${token}` }).then(response => {
+      if (response.error) {
+        if (response.data.message) {
+          dispatch(setErrorMessage(`${response.data.message}. Try resending reset password email.`))
+        } else {
+          dispatch(setErrorMessage(GENERAL_ERROR))
+        }
+        dispatch(handleGetResetFailure())
```
matt-sm: do you need this?

app/routes.js
Blackfry: hyphen?

app/actions/resetPasswordActions.js
Blackfry: It's used in the resetpasswordform to toggle pagealerts

app/pages/CreateUserPage.js
```diff
@@ -19,7 +19,10 @@ export class CreateUserPage extends BaseForm {
   }
 
   componentWillMount() {
-    const tokenString = this.props.match.params.tokenstring
+    const tokenString = this.props.location.pathname.substring(
```
matt-sm: not sure - the router match on /createuser
should you call `loadRegistrationData()` if there is no token?
and would /createuser return a 404?

PR #503, opened 2017-09-01
Airbnb

matt-sm:
- switched the full set of rules on with basically 2 exceptions:
  - jsx file extensions
  - objects as proptypes
- couple of files copied from src folder were too messed up and I found it safer to just disable rules
- some shared form components are not being used in our code and do not have tests so impossible to know if changes work
- there are still some inconsistencies in the app folder that I'll fix in a future branch


app/components/shared/form/Datefield.js
```diff
@@ -29,6 +24,12 @@ class Datefield extends React.Component {
     }
   }
 
+  state = {
```
Blackfry: https://facebook.github.io/react/docs/react-component.html#constructor.
> The constructor for a React component is called before it is mounted.
If you want to have state ready for when the component is mounted, you would set them within the `constructor` method.

matt-sm: this is the rule: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md

I don't see any example in the Facebook docs that have a separate state() function as a class member so I feel like it should be deleted

matt-sm: This is an example of a component changing with no tests that is not used in app folder

app/components/validators.js
```diff
@@ -15,9 +17,7 @@ export const required = val => {
   }
 
   if (val !== null && typeof val === 'object') {
-    return values(val).filter(function(v) {
-      return v
-    }).length
+    return values(val).filter(v => v).length
```
Blackfry: not related to your change but I think you could remove `.filter(v => v)` as it's returning 1 for 1

PR #505, opened 2017-09-05
brief response attachment upload

maxious:


matt-sm:
Need to make changes:
- base route in routes.js
- create a Page with sub routes, redux, etc
- stateless form
- lots of little eslint errors now the new rules are switched on

maxious:
Form now stateless and accessed via a Page

apps/marketplace/actions/briefActions.js
```diff
@@ -34,21 +27,16 @@ export const loadBrief = briefId => dispatch => {
 
 export const handleBriefResponseSuccess = () => ({ type: BRIEF_RESPONSE_SUCCESS })
 
-export const handleBriefResponseFailure = response => {
+export const handleBriefResponseFailure = response => dispatch => {
```
elisechant: don't put a `dispatch` here! dispatch belongs in containers. this is stinky

apps/marketplace/components/shared/form/FilesInput.js
```diff
@@ -0,0 +1,77 @@
...
+      </div>
+    )
+  }
+}
+
+const uploadDocument = (id, url, file) => () => {
+  const data = new FormData()
+  data.append(id, file)
+
+  return dmapi({
```
elisechant: @matt-sm just did an auth PR, isn't there session data? also, this might be better in a middleware?

apps/marketplace/components/shared/form/FilesInput.js
```diff
@@ -0,0 +1,77 @@
...
+  }).then(response => {
+    if (response.error) {
+      const error = new Error(response.statusText)
+      error.response = response
+      throw error
+    } else {
+      return response.data.filename
+    }
+  })
+}
```
elisechant: catch?

config/webpack.config.prod.js
```diff
@@ -114,7 +114,7 @@ module.exports = [{
     filename: '[name].[chunkhash:8].js',
     chunkFilename: '[name].chunk.js',
     // We inferred the "public path" (such as / or /my-project) from homepage.
-    publicPath: publicPath
+    publicPath: 'https://dm-frontend.apps.platform.digital.gov.au/bundle/'
```
elisechant: I would add this hardcoded path to paths.js

config/webpack.config.prod.js
```diff
@@ -230,7 +230,7 @@ module.exports = [{
     new RollbarSourceMapPlugin({
       accessToken: process.env.ROLLBAR_TOKEN || 'notoken',
       version: process.env.CIRCLE_SHA1 || 'noversion',
-      publicPath: publicPath
+      publicPath: 'https://dm-frontend.apps.platform.digital.gov.au/bundle/'
```
matt-sm: these don't look right

apps/marketplace/actions/briefActions.js
Blackfry: I think it's fine to put thunk async actions here. Primary actions in containers 👍

config/webpack.config.prod.js
maxious: turns out rollbar wants paths that start with http for a variable called "minified_url" https://rollbar.com/docs/api/sourcemaps/ but the plugin doesn't mention that https://github.com/thredup/rollbar-sourcemap-webpack-plugin/blob/master/src/RollbarSourceMapPlugin.js#L99

apps/marketplace/actions/briefActions.js
```diff
@@ -1,25 +1,18 @@
...
 
-export const handleBriefInfoFailure = response => ({
-  type: BRIEF_INFO_HAS_ERRORED,
-  // eslint-disable-next-line
-  errorMessage: response && response.data ? response.data.message : 'unknown server error'
-})
-
+export const handleBriefInfoFailure = response => dispatch => {
+  dispatch(
+    setErrorMessage(response && response.data && response.data.message ? response.data.message : 'unknown server error')
```
matt-sm: you could just have a standard message defined in `messageConstants.js`

apps/marketplace/actions/briefActions.js
matt-sm: Excuse my naivety @elisechant, but doesn't @gaearon himself say this is ok https://github.com/gaearon/redux-thunk#composition ?

config/webpack.config.prod.js
matt-sm: well at v least make it a var rather copy/pasting so many times

maxious:
fixed and rebased :)

apps/marketplace/actions/briefActions.js
```diff
@@ -34,21 +26,16 @@ export const loadBrief = briefId => dispatch => {
...
-export const handleBriefResponseFailure = response => {
+export const handleBriefResponseFailure = response => dispatch => {
   switch (response.status) {
-    case 409:
-      return {
-        type: BRIEF_RESPONSE_DUPLICATE_FAILURE,
-        // eslint-disable-next-line
-        errorMessage: response.data ? response.data.message : 'a user with this email address already exists'
-      }
+    case 409: {
```
marcoalfonso: I prefer to manage side effects through Higher order components (axios). I think it is neater

apps/marketplace/components/shared/form/FileInput.js
```diff
@@ -0,0 +1,121 @@
+import React from 'react'
+import isEmpty from 'lodash/isEmpty'
+import get from 'lodash/get'
+
+import styles from './FilesInput.scss'
+
+class FileInput extends React.Component {
+  constructor() {
+    super()
+    this.onReset = this.onReset.bind(this)
```
marcoalfonso: You don't need to bind functions in the constructor if you use es6 arrow function syntax. That autobinds the function to this.

apps/marketplace/components/shared/form/FileInput.js
```diff
@@ -0,0 +1,121 @@
...
+    this.onUpload = this.onUpload.bind(this)
+  }
+
+  state = {
+    uploading: undefined,
+    file: undefined,
+    errors: undefined
+  }
+
+  onUpload(e) {
```
marcoalfonso: const onUpload = (e) => { }

Refactor to es6 arrow function

apps/marketplace/components/shared/form/FileInput.js
```diff
@@ -0,0 +1,121 @@
...
+    e.preventDefault()
+    const { id, model, name, url, onUpload, removeDocument, updateDocumentName, createDocument } = this.props
+    const file = this.state.file
+    this.setState({
```
marcoalfonso: Can we use props instead of using internal component state?

apps/marketplace/components/shared/form/FilesInput.scss
```diff
@@ -0,0 +1,16 @@
+.bordered_list {
```
marcoalfonso: This should be css modules

apps/marketplace/pages/BriefPage.test.js
```diff
@@ -0,0 +1,27 @@
...
+  const state = {
+    userRegistrationDetails: {
+      name: 'Jeff Labowski',
+      email_address: 'e@mail.com'
+    },
+    loadCompleteSuccess: true,
+    loadRegistrationData: () => true
+  }
+
+  it('BriefPage renders without errors', () => {
```
matt-sm: I think you'll find an 'it' in a 'test' doesn't actually run

apps/marketplace/pages/BriefPage.js
```diff
@@ -0,0 +1,83 @@
...
+    const { brief } = this.props
+    this.props.handleBriefResponseSubmit(brief.id, values)
+  }
+
+  render() {
+    const { match } = this.props
+    return (
+      <div className="brief-page">
+        <Switch>
+          <Route
```
matt-sm: This should now be `<PrivateRoute ..>` it will ensure they are prompted to login if not already

apps/marketplace/actions/briefActions.js
```diff
@@ -1,25 +1,17 @@
...
 })
 
-export const handleBriefInfoFailure = response => ({
-  type: BRIEF_INFO_HAS_ERRORED,
-  // eslint-disable-next-line
-  errorMessage: response && response.data ? response.data.message : 'unknown server error'
-})
-
+export const handleBriefInfoFailure = response => dispatch => {
+  dispatch(setErrorMessage(response && response.data && response.data.message ? response.data.message : GENERAL_ERROR))
```
matt-sm: don't we control whats in the response?  could you not ensure there's a message in the api?

apps/marketplace/pages/BriefPage.js
```diff
@@ -0,0 +1,83 @@
...
+            path={`${match.url}/respond`}
+            render={() =>
+              <BriefResponseForm
```
matt-sm: if the initial data load fails, we don't want to load the form if you see `CreateUserPage` it shows an `ErrorBox` in this case

apps/marketplace/pages/BriefPage.js
matt-sm: We're also currently do in all the permission checks on the post - but could be misleading to a user that fills out a response thinking that they are allowed to

apps/marketplace/components/shared/form/FileInput.js
```diff
@@ -0,0 +1,114 @@
...
+      <div key={fileField} className="callout-no-margin">
+        <div>
+          {errors && <span className="validation-message">There was an error uploading the file</span>}
+
+          {isEmpty(doc) &&
+            !this.state.uploading &&
+            !this.state.file &&
+            <div id={fileField}>
+              <p>
+                <input type="file" id={fileField} name={fileField} accept=".pdf,.odt" onChange={this.onChange} />
```
matt-sm: needs some styling:

screen shot 2017-09-19 at 2 27 44 pm

apps/marketplace/components/shared/form/FileInput.js
```diff
@@ -0,0 +1,114 @@
...
+                    <a href="#delete" onClick={this.onReset}>
+                      Delete
+                    </a>
+                  </div>
+                </li>
+              </ul>
+            </div>}
+        </div>
+
+        {this.state.uploading && <p>Uploading...</p>}
```
matt-sm: if there is an upload error, seems to get stuck in this state

apps/marketplace/components/Brief/BriefResponseForm.js
```diff
@@ -0,0 +1,179 @@
...
+              messages={{
+                required: 'A contact email is required',
+                validEmail: 'A valid contact email is required'
+              }}
+            />
+            {supplierCode &&
+              brief.lotSlug &&
+              brief.lotSlug === 'digital-professionals' &&
+              <span>
+                <Textfield
```
matt-sm: I typed `$100` here - client didn't complain but server returned `not_money_format` (which was not displayed in error box)

apps/marketplace/components/Brief/BriefResponseForm.js
```diff
@@ -0,0 +1,179 @@
...
+      e.focus()
+    }
+  }
+
+  return (
+    <div className="row">
+      <div className="col-sm-push-2 col-sm-8 col-xs-12">
+        <article role="main">
+          {briefResponseSuccess
+            ? <PageAlert as="success">
```
matt-sm: I can't get this to display - I can see `briefResponseSuccess` is set to true and the response is created but no success message displayed

apps/marketplace/components/Brief/BriefResponseForm.js
```diff
@@ -0,0 +1,179 @@
...
+              </span>}
+
+            <p className="uikit-page-alerts uikit-page-alerts--info">
+              <b>Remember</b>
+              <br />
+              — Only one 1 response per seller is allowed<br />
+              — You cannot edit your application after submitting<br />
+            </p>
+            <p>
+              <input className="uikit-btn" type="submit" value="Submit" onClick={submitClicked} />
```
matt-sm: need to add the spinner button

apps/marketplace/actions/briefActions.js
maxious: what about a network error (ie. when I forget to start API ;) )

Blackfry: we can catch this case in the api client (see my slack msg):

```
const sanitiseErrorResponse = data => {
  if (!data || data.slice(0, 2) === "<?") { // when unhandled api errors return xml
    return GENERAL_ERROR
  } else {
    return data
  }
}

...
.catch(error => {
      if (error.response) {
         return {
             ...
             errorMessage: sanitiseErrorResponse(error.response.data), // data is a string

```
then handleFailure can be:
~~dispatch(handleBriefInfoFailure(response))~~
```
dispatch(setErrorMessage(response.errorMessage))
```

apps/marketplace/components/Brief/BriefResponseForm.js
```diff
@@ -0,0 +1,170 @@
+/* eslint-disable react/no-array-index-key */
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Form } from 'react-redux-form'
+import { Redirect } from 'react-router-dom'
+
+import { required, validEmail, validPercentage } from '../../components/validators'
```
matt-sm: We're now converting all these to absolute paths

apps/marketplace/pages/BriefPage.js
```diff
@@ -0,0 +1,112 @@
...
+            component={BriefResponseSubmitted}
+            setFocus={setFocus}
+            {...this.props}
+          />
+          <Route
+            path={`${match.url}/respond`}
+            render={() =>
+              <span>
+                {loadBriefSuccess
+                  ? <BriefResponseForm
```
matt-sm: This also be a `PrivateRoute`

matt-sm: In fact you could for now make the entire base brief route private

apps/marketplace/components/PrivateRoute.js
```diff
@@ -10,7 +10,7 @@ const PrivateRouteComponent = props => {
       {...rest}
       render={values =>
         loggedIn
-          ? <Component {...values} />
+          ? <Component {...values} {...props} />
```
Blackfry: looks like the component is receiving itself as props 🤔. `...rest` should be enough

apps/marketplace/pages/BriefPage.test.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import renderer from 'react-test-renderer'
+import expect from 'expect'
+import { shallow } from 'enzyme'
+import { BriefPage } from './BriefPage'
+
+test('Test suite for BriefPage page', () => {
```
Blackfry: this test won't run, replace top level `test` with `describe`

apps/marketplace/components/shared/form/FileInput.js
maxious: restyling

PR #510, opened 2017-09-07
Report page v2

Blackfry:
move app to apps


apps/marketplace/components/Reports/ReportItem.js
```diff
@@ -0,0 +1,64 @@
+// will swap this with <img>
```
maxious: misplaced comment?

apps/marketplace/components/Reports/ReportItem.js
```diff
@@ -0,0 +1,64 @@
...
+/* eslint-disable react/no-danger  */
+/* eslint-disable import/no-dynamic-require */
```
maxious: is this still required?

Blackfry: I will swap `dangerouslyset` for an `img` but for now eslint complains

PR #511, opened 2017-09-08
prod json-loader

Blackfry:


config/webpack.config.prod.js
```diff
@@ -65,7 +65,7 @@ const loaders = [
   // allow it implicitly so we also enable it.
   {
     test: /\.json$/,
-    loader: 'json'
+    loader: 'json-loader'
```
matt-sm: remember to remove this

package.json
```diff
@@ -89,6 +88,7 @@
     "express": "4.15.3",
     "fast-json-patch": "^2.0.4",
     "history": "^4.6.3",
+    "json-loader": "0.5.4",
```
matt-sm: assume we also don't need this

PR #515, opened 2017-09-11
Feature/orams homepage

marcoalfonso:


apps/orams/components/Header/Header.js
```diff
@@ -0,0 +1,35 @@
...
+          <div className={styles.userNav}>
+            <div id="react-bundle-auth-header-state" />
+            <div id="react-bundle-auth-header">
+              <ul data-reactroot="" id="main-navigation" className={styles.inlineLinks}>
+                <li>
+                  {this.props.memberInfo.isAuthenticated
+                    ? <span>
+                        {this.dashBoardLink()}
+                      </span>
+                    : <a href="/2/signup">Join the Marketplace</a>}
```
matt-sm: that "2" value needs to come from a constant.  I suggest rather than `routes.js` put it in `constants.js`.

PR #518, opened 2017-09-13
Absolute paths

marcoalfonso:


apps/orams/index.js
```diff
@@ -3,9 +3,9 @@ import ReactDOM from 'react-dom'
 import { Provider } from 'react-redux'
 import { BrowserRouter } from 'react-router-dom'
 
-import HomepageHeader from './components/HomepageHeader/HomepageHeader'
-import Banner from '../shared/Banner/Banner'
-import Footer from '../shared/Footer/PageFooter'
+import Banner from 'shared/Banner/Banner'
```
matt-sm: we should still follow import format from style guide:
- absolute imports
- blank line
- relative imports
- blank line
- css modules

apps/orams/components/HomePage/HomePage.js
```diff
@@ -1,6 +1,26 @@
...
+            <a href="/become-a-seller">Become a seller</a>
+          </h2>
+          <p>Access more digital opportunities across government.</p>
+        </div>
+      </div>
+    </section>
+    <section className={styles.contentMain} />
+  </div>
 
 HomePage.defaultProps = {}
```
matt-sm: if there are no props, do we need to worry about props?

marcoalfonso: Fixed

PR #519, opened 2017-09-14
Refactor signup page

Blackfry:


apps/marketplace/components/SignupForm/SignupForm.js
```diff
@@ -1,371 +1,299 @@
 /* eslint-disable camelcase */
 import React from 'react'
 import PropTypes from 'prop-types'
-import { connect } from 'react-redux'
 import { Form } from 'react-redux-form'
 import PageAlert from '@gov.au/page-alerts'
 
 import { required, validEmail, governmentEmail } from '../../components/validators'
 import Layout from '../../components/shared/Layout'
```
matt-sm: we shouldn't put all the extra blank lines in the imports:
- absolute
- blank line
- relative
- blank line
- css modules

apps/marketplace/components/SignupForm/SignupForm.js
```diff
@@ -1,371 +1,299 @@
...
-  signupSuccess: state.user.signupSuccess,
-  currentlySending: state.app.currentlySending
-})
-
-const mapDispatchToProps = dispatch => ({
-  handleSignupSubmit: model => dispatch(handleSignupSubmit(model))
-})
+SignupForm.defaultProps = {
+  csrf_token: '',
+  signupSuccess: null
```
matt-sm: should that be `false`

apps/marketplace/components/SignupForm/SignupForm.js
```diff
@@ -1,371 +1,299 @@
...
+            </Layout>}
         </div>
       </div>
-    )
-  }
+    </div>
+  )
 }
 
 SignupForm.propTypes = {
```
matt-sm: I think your missing `submitClicked`

apps/marketplace/components/SignupForm/SignupForm.js
Blackfry: I normally use null for defaults so if it is false I know it's been set that way from an action

PR #533, opened 2017-09-29
show a loading indicator when ping has not returned

matt-sm:
Also:
- took PrivateRoute out of shared and renamed files
- fixed the page reload on edit profile


apps/marketplace/MarketplaceAuthenticatedRoute.js
```diff
@@ -0,0 +1,43 @@
...
+            />
+      }}
+    />
+  )
+}
+
+PrivateRouteComponent.propTypes = {
+  component: PropTypes.func.isRequired,
+  loggedIn: PropTypes.bool.isRequired
+}
```
Blackfry: add `customRedirectPath, currentlySending` to proptypes

PR #535, opened 2017-10-04
Seller catalogue page

marcoalfonso:


views/orams.ejs
```diff
@@ -1,19 +1,31 @@
...
+<!--[if !(IE)]><!-->
+  <html class="js" lang="en">
+<!--<![endif]-->
 	<head>
     <title>ORAMS Marketplace</title>
 		<meta charSet="utf-8"/>
     <meta name="viewport" content="width=device-width, initial-scale=1"/>
 		<link rel="shortcut icon" href="/static/media/favicon.png" type="image/x-icon" />
-
+    <link rel="stylesheet" type="text/css" href="/bundle/custom.css"/>
```
maxious: Changing order would make it harder to override uikit?

marcoalfonso: Good catch. Let me fix

PR #542, opened 2017-10-10
Feature/orams footer header

marcoalfonso:


apps/orams/components/Footer/PageFooter.js
```diff
@@ -0,0 +1,45 @@
+/* eslint-disable  react/no-danger */
```
maxious: do we need dangerous use of HTML?

apps/orams/components/HomePage/HomePage.js
```diff
@@ -1,9 +1,10 @@
+/* eslint-disable */
 import React from 'react'
 import styles from './HomePage.scss'
 
 const HomePage = () =>
   <div>
-    <section className={styles.heroMain}>
+    {/*<section className={styles.heroMain}>
```
matt-sm: delete this?

PR #544, opened 2017-10-17
Link to new brief submitted screen from dashboard

maxious:


public/custom.css
```diff
@@ -153,6 +144,18 @@ form .hint {
 input[type="radio"].uikit-control-input__input:focus + .uikit-control-input__text:before {
   border-radius: 15px;
 }
+.uikit-btn {
```
matt-sm: @marcoalfonso - thoughts?

maxious: This should restore the styling of "buttons wrongly marked up :)" http://uikit.apps.staging.digital.gov.au/packages/buttons/tests/site/ without changing every link in every article.

public/custom.css
```diff
@@ -153,6 +144,18 @@ form .hint {
...
+  color: #fff !important;
+}
+.uikit-btn--tertiary {
+  color: #313131 !important;
```
marcoalfonso: Don't use important!!!  

Just put a class on top and do 
.class .uikit-btn

public/custom.css
marcoalfonso: NO IMPORTANT!

public/custom.css
```diff
@@ -153,6 +144,18 @@ form .hint {
...
+}
+.uikit-btn:focus,.uikit-btn:hover {
+  background-color: #135566 !important;
```
marcoalfonso: 👎

maxious:
now with less !important @marcoalfonso

PR #550, opened 2017-10-19
Add frequent mistakes callout to case studies

maxious:


src/bundles/CaseStudy/components/CaseStudyForm/AssessmentCriteria.js
```diff
@@ -0,0 +1,368 @@
+import React from 'react'
+
+const AssessmentCriteria = (props) => {
+  const {service_slug} = props
+  switch (service_slug) {
```
matt-sm: something about a huge `switch` that doesn't feel right...

PR #551, opened 2017-10-20
Reenable report page with accessible captions for long text

maxious:


apps/marketplace/components/Reports/ReportItem.js
```diff
@@ -0,0 +1,104 @@
+/* eslint-disable jsx-a11y/no-noninteractive-tabindex */
+import React from 'react'
+import PropTypes from 'prop-types'
+import cx from 'classnames'
```
marcoalfonso: Why name classnames cx and not classnames? This is confusing.

apps/marketplace/components/Reports/ReportItem.js
```diff
@@ -0,0 +1,104 @@
...
+
+  return (
+    <div className="col-sm-12 report-item">
+      <div className={styles.reportItem}>
+        {heading &&
+          <h2 className={`${styles.reportItemHeading} uikit-display-2`}>
+            {heading}
+          </h2>}
+        {subitems.map((subitem, id = uniqueID()) =>
+          <p className={`${styles.reportSubitemContainer} ${subitem.fullWidth ? 'col-sm-12' : 'col-sm-6'}`} key={id}>
```
marcoalfonso: why are you controlling responsiveness with a prop? This should be done with CSS, we had this discussion with Mark, responsiveness has to be done with css!

apps/marketplace/components/Reports/ReportItem.js
```diff
@@ -0,0 +1,104 @@
...
+            {subitem.image &&
+              <span
+                className={computeClassname(
+                  styles.reportItemImage,
+                  styles.reportFormattedQuoteImage,
+                  subitem.formattedImage
+                )}
+              >
+                {subitem.imageCaption
+                  ? <figure
```
marcoalfonso: why use the figure html element? It is not supported in IE.

apps/marketplace/components/Reports/ReportItem.js
```diff
@@ -0,0 +1,104 @@
...
+                        alt={subitem.imageAlt}
+                      />
+                      <figcaption
+                        className={styles.sr_only}
+                        id={`${subitem.image.replace('768-', '').replace('.svg', '')}-figcaption`}
+                      >
+                        {subitem.imageCaption}
+                      </figcaption>
+                    </figure>
+                  : <img
```
marcoalfonso: We should avoid using ternarie, it is better to just render a section above the render(). You can do something {renderThisSection()} and then on top you can just use if statements to return blocks.

apps/marketplace/components/Reports/ReportItem.js
maxious: Agreed - this PR was just undeleting that code but I've fixed this and other issues now.

PR #552, opened 2017-10-23
 Switch role=button to class=button

maxious:


matt-sm:
this was raised in the accessibility review?

maxious:
yes while this visually looks like a button it should be interacted with as a link and acts like a link so removing this role was suggested.

PR #556, opened 2017-10-25
Add more common application mistake callouts

maxious:


src/bundles/CaseStudy/components/CaseStudyForm/CaseStudyForm.js
```diff
@@ -94,7 +92,7 @@ class CaseStudyForm extends BaseForm {
                 </p>
               </div>    
             )}
-          <div className={styles.calloutMistake}>
+          <div className="calloutMistake">
```
marcoalfonso: Why are you using a class for this and why not a css module? Where is this class defined?

maxious: Unfortunately CSS for this old application process loads from a python app: https://github.com/AusDTO/dto-digitalmarketplace-supplier-frontend/commit/916af5b646159c4d1d30945a8ea6188d31c5f4a7

PR #557, opened 2017-10-29
Remove inline styles from newer marketplace react views

maxious:


apps/marketplace/components/Onboarding/OnboardingContainer.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { withRouter, Switch, Route } from 'react-router-dom'
-import NotFound from '../shared/NotFound'
+import NotFound from '../NotFound'
```
matt-sm: should probably make all these relative to root

PR #559, opened 2017-10-31
Display a warning message when javascript is disabled

maxious:


views/index.ejs
```diff
@@ -1,50 +1,58 @@
...
+  </script>
+
+</head>
+<body class="uikit-grid uikit-body">
+<noscript class="col-sm-push-2 col-sm-8 col-xs-12">
+  <div class="uikit-page-alerts uikit-page-alerts--error" role="alert">
+    <h4 id="validation-masthead-heading" tabindex="-1"> Digital Marketplace requires Javascript and Cookies running in your browser.</h4>
+    To continue you can:
+    <ul>
+      <li>Enable Javascript (<a href="https://www.enable-javascript.com/" target="_blank">how do I enable Javascript?</a>) or Cookies (<a href="http://www.whatarecookies.com/enable.asp">how do I enable Cookies?</a>)</li>
```
matt-sm: or Cookies?  Don't they just need to enable js?

maxious: Logging in also requires cookies for session handling so it was suggested to include that as a troubleshooting step too.

matt-sm: shouldn't it read "and Cookies" then?

maxious: Makes sense to me and matches the heading just above. Updated.

PR #561, opened 2017-11-06
Fix case study validation to halt progress

maxious:


src/bundles/SellerRegistration/components/Signup/Signup.js
```diff
@@ -112,6 +112,9 @@ class Signup extends React.Component {
       actions.stepNextPersist(this.nextStep.pattern, this.step);
     },
     onSubmitFailed: (e) => {
+      if (e && 'preventDefault' in e) {
```
matt-sm: would there be a case where e is not an event?

maxious: I wonder if it ever is an event... https://github.com/davidkpiano/react-redux-form/blob/8e8a375f02c1a978bf1804b2cc5a5413e96bf1b4/src/components/form-component.js#L296

matt-sm: If it's not, do we need this?

PR #570, opened 2017-11-14
collaborate updates

maxious:


apps/marketplace/components/Collaborate/Landing.js
```diff
@@ -0,0 +1,298 @@
...
+          <strong className={styles.pop}>
+            <font color="#870f48">
+              {52}
+            </font>
+          </strong>{' '}
+          <article>funded projects</article>
+        </div>
+      </div>
+    </div>
+    <article className={`${styles.filters} row`}>
```
marcoalfonso: Why two rows? needs a column after row

apps/marketplace/components/Collaborate/Landing.js
```diff
@@ -0,0 +1,298 @@
...
+                      </div>
+                      <br />
+                      <p>
+                        {val.description}
+                      </p>
+                      <hr />
+                    </div>
+                  )
+                }
+                return undefined
```
marcoalfonso: Is this necesary?

maxious: eslint wants returns
  109:31  error  Expected to return a value at the end of arrow function  array-callback-return
  109:31  error  Expected to return a value at the end of arrow function  consistent-return

apps/marketplace/components/Collaborate/Landing.js
maxious: or no rows, updated.

PR #572, opened 2017-11-16
Buyer profile

marcoalfonso:


apps/orams/components/Profile/Profile.js
```diff
@@ -0,0 +1,183 @@
...
+    this.state = {}
+  }
+
+  componentDidMount() {
+    this.props.loadProfileData()
+  }
+
+  render() {
+    const { profileData } = this.props
+    console.log('profileData', profileData)
```
maxious: is log needed?

apps/orams/components/Profile/Profile.js
```diff
@@ -0,0 +1,183 @@
+/* eslint-disable */
```
matt-sm: aren't Profile and SellerProfile essentially the same component?

apps/orams/pages/ProfilePage.js
```diff
@@ -0,0 +1,42 @@
+/* eslint-disable */
```
matt-sm: You could also refuse the other seller profile page

PR #577, opened 2017-11-22
change to latest api routes

matt-sm:


matt-sm:
Good to see you guys keeping up with the latest version 🥇

PR #579, opened 2017-11-24
Switch to autopilot, use user-provided-services

aeijdenberg:
Not sure whether this will work, given that it's invoking Python in a nodejs buildpack, but figured trying it would be a fine way to test. :)

### DIFF
--- a/scripts/ups_as_envs.py
+++ b/scripts/ups_as_envs.py
@@ -1,0 +1,9 @@
+#!/usr/bin/env python
+
+import os
+import json
+import sys
+
+for ups in json.loads(os.environ.get("VCAP_SERVICES", "{}")).get("user-provided", []):
+    for k, v in ups.get("credentials", {}).items():
+        sys.stdout.write("export %s='%s'\n" % (k, v.replace("'", "'\\''")))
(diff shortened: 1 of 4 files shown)


scripts/ups_as_envs.py
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env python
+
+import os
+import json
+import sys
+
+for ups in json.loads(os.environ.get("VCAP_SERVICES", "{}")).get("user-provided", []):
+    for k, v in ups.get("credentials", {}).items():
+        sys.stdout.write("export %s='%s'\n" % (k, v.replace("'", "'\\''")))
```
matt-sm: why `eval` these `stdout.write` instead of `os.environ[] = ""`

aeijdenberg: Pretty sure that modifying `os.environ[]` will only modify the env variables for this process, not it's parent, which is the one that launches the actual server.

So for example that might work if we put this script inside of the launched app, but I kind of like it external - as it will work across non-Python apps that way.

(and the man page for modifying `os.environ[]` had enough caveats and scary warnings that it scared me off: <https://docs.python.org/2/library/os.html#process-parameters>)

matt-sm: Makes sense

PR #583, opened 2017-11-29
Several hot fixes

marcoalfonso:


apps/orams/pages/LoginPage.js
```diff
@@ -46,7 +46,7 @@ export class LoginPageComponent extends BaseForm {
                 model={model}
                 currentlySending={currentlySending}
                 rootPath={rootPath}
-                framework="orams"
+                framework="ORAMS"
```
matt-sm: not sure about this one - if this is value sent to api it needs to be lower case

marcoalfonso: Didn't break things, can we merge this?

PR #589, opened 2017-12-04
Final step

marcoalfonso:


apps/orams/actions/appActions.js
```diff
@@ -137,3 +145,47 @@ export const loadInvitation = token => dispatch => {
...
+export const loadUserToCreate = token => dispatch => {
+  dispatch(sendingRequest(true))
+  dmapi({
+    method: 'get',
+    url: `/tokens/${token}`
+  }).then(response => {
+    if (response.error) {
+      dispatch(setErrorMessage(GENERAL_ERROR))
+    } else {
+      console.log("RESPONSE", response.data)
```
matt-sm: remove `console.log` ?

apps/orams/components/CreatePassword/CreatePassword.js
```diff
@@ -0,0 +1,112 @@
...
+  let hasFocused = false
+  const setFocus = e => {
+    if (!hasFocused) {
+      hasFocused = true
+      e.focus()
+    }
+  }
+
+  return (
+    <main>
```
matt-sm: when  the token is still being decoded, shouldn't the user see a "loading" indicator?

marcoalfonso: Ok put a loading indicator. If the token is invalid... mmmh it shouldn't be invalid :P

PR #592, opened 2017-12-11
Feature/authentication framework changes

marcoalfonso:


apps/marketplace/actions/appActions.js
```diff
@@ -51,12 +52,20 @@ export function setAuthState(newState) {
...
+}
+
 export const fetchAuth = () => dispatch => {
   dispatch(sendingRequest(true))
+  dispatch(setAuthFrameworkError(false))
   dmapi({ url: '/ping' }).then(response => {
     if (response.error) {
       dispatch(setErrorMessage(GENERAL_ERROR))
     } else {
+      if (response.data.framework === 'orams') {
```
matt-sm: safer to do !== 'digital-marketplace' - we could fix it if we introduce a 3rd app

PR #607, opened 2018-01-25
Update BriefSubmitted.js

nellied:
Replaced button style with a link, updated copy and removed some break tags.


apps/marketplace/components/Brief/BriefSubmitted.js
```diff
@@ -20,31 +20,30 @@ const BriefSubmitted = props =>
                   until {format(new Date(props.brief.applicationsClosedAt), 'MMMM Do, YYYY [at] ha [in Canberra]')}
                 </span>}
             </p>
-            <a className="uikit-btn" href={`/${props.brief.frameworkSlug}/opportunities/${props.brief.id}`}>
-              View live brief
-            </a>
+            <a href={`/${props.brief.frameworkSlug}/opportunities/${props.brief.id}`}>View live opportunity</a>
```
marcoalfonso: Why did you remove the className="uikit-btn" ?

malmas18: Neil wanted to save some vertical space.

So the class was removed in favour of a normal link.

#MAR-1429

PR #615, opened 2018-02-12
MAR-973 add defaults to name and email_address to prevent null being …

malmas18:
…sent


apps/marketplace/components/Feedback/Feedback.js
```diff
@@ -28,9 +28,19 @@ export class Feedback extends React.Component {
               <h4>Your feedback has been sent, thank you for helping us improve the Marketplace.</h4>
             </PageAlert>
           : <LocalForm onSubmit={val => this.props.handleSubmit(val)}>
-              <Control.input type="hidden" model=".commentQuestion" defaultValue={this.props.commentQuestion} />
-              <Control.input type="hidden" model=".difficultyQuestion" defaultValue={this.props.difficultyQuestion} />
-              <Control.input type="hidden" model=".objectAction" defaultValue={this.props.objectAction} />
+              <Control.input
```
matt-sm: what are these changes for?

malmas18: to fix the same error.  added value="" to it.  
the formatting was required by eslint.

PR #639, opened 2018-04-02
Multi Candidate Frontend

matt-sm:
This is based of Marco's original branch.  The following are out of scope:
- withdraw
- edit application

Some of the other highlights:
- both the old and new brief pages for specialists need to keep working for now
- outcomes are not affected
- did some testing on screenreader and have added a live region to the app root
- have added tests for new components


apps/marketplace/components/Brief/BriefSpecialistResponseForm.js
```diff
@@ -0,0 +1,247 @@
...
+  specialistName,
+  specialistNumber,
+  addAnotherClicked,
+  addAnotherSpecialist
+}) =>
+  <div className="row">
+    <DocumentTitle title="Brief Response - Digital Marketplace">
+      <div className="col-sm-push-2 col-sm-8 col-xs-12">
+        <article role="main">
+          {(briefResponseSuccess && !addAnotherSpecialist) || briefResponses.length === 3
```
malmas18: it would be safer to check for briefResponses.length >== 3

malmas18: put 3 in a constant?

PR #640, opened 2018-04-04
Convert to puppeteer from phantomjs

malmas18:
Use google chrome project for scraping https://github.com/GoogleChrome/puppeteer


matt-sm:
Why'd you change from phantom?

PR #648, opened 2018-04-16
Mar 1517 seller unsuccessful notify

ndavison:


apps/marketplace/components/SellerNotify/SellerNotifyEmailUnsuccessful.js
```diff
@@ -0,0 +1,30 @@
...
+      <li>
+        <em>
+          {props.brief.organisation} found answers more convincing when they were backed up with examples of recent
+          real-life scenarios.
+        </em>
+      </li>
+    </ul>
+    <p>To assist you with your future applications, the Marketplace team have put together a debrief for you.</p>
+  </div>
+
```
matt-sm: add `propTypes`?

apps/marketplace/components/SellerNotify/SellerNotifyIntroduction.js
```diff
@@ -0,0 +1,46 @@
...
+
+  handleContinueClick(e) {
+    e.preventDefault()
+    this.props.setStageStatus('introduction', 'done')
+    this.props.setStageDoneStatus('introduction', true)
+    this.props.moveToNextStage('introduction')
+  }
+
+  render() {
+    if (this.props.flow === 'unsuccessful') {
```
matt-sm: was `flow` really required for first version?  (not saying we remove it, just feels premature)

apps/marketplace/components/SellerNotify/SellerNotify.scss
```diff
@@ -0,0 +1,19 @@
+.reviewForm label {
+  font-weight: 700;
+  display: block;
+  margin-bottom: 5px;
+}
+
+.reviewForm input, .reviewForm div[contenteditable=true] {
+  border: 2px solid #bebebe;
+  width: 100%;
+  padding: 10px;
```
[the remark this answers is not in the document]
ndavison: Possibly, px is more widely supported though.

I couldn't seem to get the uikit text-inputs working, but I believe the issue will be fixed with a uikit upgrade so I think this is a good candidate to be removed and replaced once that is in place.

apps/marketplace/components/SellerNotify/SellerNotifyIntroduction.js
ndavison: It is premature, but I made the page and components to be generic enough that future flows can be added with ease given we are expecting a few, vs having to refactor for it when they come.

apps/marketplace/pages/SellerNotifyPage.js
```diff
@@ -0,0 +1,263 @@
...
+
+  moveToStage(stage) {
+    if (this.state.stages[stage]) {
+      this.props.history.push(
+        `${rootPath}/brief/${this.props.match.params.briefId}/seller-${this.props.match.params.flow}/${stage}`
+      )
+    }
+  }
+
+  selectSeller(supplierCode, callback) {
```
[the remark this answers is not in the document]
ndavison: It's my understanding redux should be used where propagating stage changes would involve intermediate components that do nothing but pass the state down to deeper child components. In this case selecting a seller (and other similar functions) is only needed by direct children of the page component so it may be overkill to involve redux. Currently the child components do not need to be connected to the redux store.

PR #651, opened 2018-05-03
Mar 1297 download resumes

ndavison:


apps/marketplace/components/Brief/BriefDownloadDocuments.js
```diff
@@ -0,0 +1,61 @@
...
+    )
+  }
+
+  render() {
+    return (
+      <span>
+        <strong>
+          Downloading documents for brief &quot;{this.props.brief.title}&quot;...{' '}
+        </strong>
+        {this.props.briefDocumentsData ? this.createDownloadElement() : <LoadingIndicatorFullPage />}
```
matt-sm: Looks a bit odd have the "Downloading.." text with the loading indicator

apps/marketplace/components/Brief/BriefDownloadDocuments.js
```diff
@@ -0,0 +1,61 @@
...
+      </span>
+    )
+  }
+}
+
+BriefDownloadDocuments.propTypes = {
+  brief: PropTypes.object.isRequired,
+  downloadBriefDocuments: PropTypes.func.isRequired
+}
```
matt-sm: `briefDocumentsData` missing?

apps/marketplace/components/Brief/BriefDownloadDocuments.js
```diff
@@ -0,0 +1,66 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import LoadingIndicatorFullPage from 'shared/LoadingIndicatorFullPage/LoadingIndicatorFullPage'
+import styles from './BriefDownloadDocuments.scss'
+
+const canDownload = () => {
+  const hasDownload = typeof document.createElement('a').download !== 'undefined'
+  const hasObjectURL = typeof window.URL !== 'undefined' && window.URL.createObjectURL !== 'undefined'
```
marcoalfonso: Don't use the window object, that is vanilla javascript and not the react way. If you want get the url you can use the match prop. Just pass the props to the component and then you can call match.url

apps/marketplace/components/Brief/BriefDownloadDocuments.js
```diff
@@ -0,0 +1,66 @@
...
+  componentDidMount() {
+    this.props.downloadBriefDocuments(this.props.brief.id)
+  }
+
+  createDownloadElement() {
+    if (canDownload()) {
+      return (
+        <a
+          className={styles.hidden}
+          href={window.URL.createObjectURL(this.props.briefDocumentsData)}
```
marcoalfonso: Also use match not window. Anti react pattern

apps/marketplace/components/Brief/BriefDownloadDocuments.js
ndavison: @marcoalfonso this is to use the Object URL API, it's not accessing the URL of the page (first it's a feature compatibility check, and the next is the actual usage). This is a browser API - is there a better way to bring it in? I may be able to reference it simply as `URL` but either way it's just as global as `window.URL`.

apps/marketplace/components/Brief/BriefDownloadDocuments.js
ndavison: Same as above.

PR #653, opened 2018-05-04
[Brief Overview] Build new overview page

tobiasbrown:
**Notes**
* Status badges have been dropped for now


apps/marketplace/components/Brief/Overview/BriefOverviewSectionLinkList.js
```diff
@@ -0,0 +1,26 @@
+import React, { Component } from 'react'
+
+import BriefOverviewSectionLinkListItem from './BriefOverviewSectionLinkListItem'
+
+import styles from './BriefOverviewSectionLinkList.scss'
+
+export class BriefOverviewSectionLinkListComponent extends Component {
```
matt-sm: this can be SFC

config/webpack.config.dev.js
```diff
@@ -282,9 +282,7 @@ module.exports = [{
         include: [
           paths.marketplaceSrc,
           paths.sharedComponents,
-          paths.appNodeModules + '/@gov.au/footer',
-          paths.appNodeModules + '/@gov.au/page-alerts',
-          paths.appNodeModules + '/@gov.au/control-input',
+          paths.appNodeModules + '/@gov.au'
```
matt-sm: you don't want to mess with webpack config

apps/marketplace/components/Brief/Overview/BriefOverviewSectionLinkList.js
matt-sm: Feels like this and parent could be combined?

config/webpack.config.dev.js
tobiasbrown: This prevents ui-kit from overwriting the styles in the header.

apps/marketplace/components/Brief/Overview/BriefOverviewSectionLinkList.js
tobiasbrown: I've merged the `LinkList` and `LinkListItem` and changed to SFC.  Added prop types too.

config/webpack.config.dev.js
matt-sm: The new uikit has been merged now so perhaps rebase from master?

apps/marketplace/routes.js
```diff
@@ -19,6 +20,7 @@ export const Routes = () =>
     <Route exact path={rootPath} component={SignupPage} />
     <Route path={`${rootPath}/signup`} component={SignupPage} />
     <Route path={`${rootPath}/create-user`} component={CreateUserPage} />
+    <PrivateRoute path={`${rootPath}/brief/:briefId/overview`} component={BriefOverviewPage} />
```
matt-sm: I would put this route in `BriefPage`

apps/marketplace/components/Brief/Overview/BriefOverview.js
```diff
@@ -0,0 +1,46 @@
...
+    super(props)
+    this.state = {}
+  }
+
+  render() {
+    const { sections, title } = this.props
+
+    return (
+      <div className="row">
+        <div className="col-md-12 col-sm-12">
```
ndavison: I believe uikit grid is like the bootstrap grid, so just `col-xs-12` here would suffice (full width on extra small, and all breakpoints above it).

apps/marketplace/components/Brief/Overview/BriefOverviewSectionList.js
```diff
@@ -0,0 +1,68 @@
...
+    }
+  }
+
+  preventDefault = e => {
+    e.preventDefault()
+  }
+
+  toggle = () => {
+    this.setState({
+      showLinks: !this.state.showLinks
```
ndavison: This one looks pretty benign but in general when setting React state you should not reference the current state using `this.state` - this is because state changes can be asynchronous so two quick fires of this may not be handled properly. The recommendation is to pass `setState` a function, which is given the previous state as an arg, e.g.:

```
this.setState(prevState => ({
  showLinks: !prevState. showLinks
}));
```

Further reading: https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

config/webpack.config.dev.js
ndavison: this will probably be a merge conflict when rebased from master, but I think this approach is better as it means we don't have to modify the webpack config every time we add a new uikit component to the code base.

FYI the `include` is a bit misleading. It doesn't mean webpack will include this path, it is just used in conjunction with `test` to match imports to a particular loader (in this case, `babel`). So I don't believe there is any efficiency found in specifying uikit components, vs just matching on the entire `@gov.au` npm scoped packages.

https://webpack.js.org/configuration/module/#condition (we're very far behind in webpack version but I believe this doco is accurate for our version too)

apps/marketplace/reducers/briefReducers.js
```diff
@@ -20,13 +21,25 @@ const defaultBriefState = {
...
   switch (action.type) {
+    case BRIEF_OVERVIEW_SUCCESS:
+      return {
+        ...state,
+        framework: action.framework,
+        loadBriefOverviewSuccess: true,
+        loadBriefOverviewErrored: false,
+        lot: action.lot,
+        sections: action.sections,
+        title: action.title
```
ndavison: would it be nicer to scope all these to a `overview`  attribute? so components access it in the state via `state.brief.overview.title` vs `state.brief.title`. It's probably safe not to worry but it may help give context to the state values when referenced in a connected component.

apps/marketplace/components/Brief/Overview/BriefOverview.js
tobiasbrown: `col-xs-12`added.

apps/marketplace/components/Brief/Overview/BriefOverviewSectionList.js
tobiasbrown: Thank you.  This has been updated so that `setState` is passed a function.

apps/marketplace/components/Brief/Overview/BriefOverviewSectionList.js
```diff
@@ -0,0 +1,68 @@
...
+    this.setState(prevState => ({
+      showLinks: !prevState.showLinks
+    }))
+  }
+
+  buildLinkListItem = link => {
+    const item = {
+      li: {
+        className: 'linkListItem',
+        'data-complete': link.complete
```
matt-sm: is this accessible?  using a screen reader, would you know the item was completed?

apps/marketplace/components/Brief/Overview/BriefOverviewSectionList.js
```diff
@@ -0,0 +1,68 @@
...
+
+export class BriefOverviewSectionListComponent extends Component {
+  static propTypes = {
+    links: PropTypes.array.isRequired
+  }
+
+  constructor(props) {
+    super(props)
+    this.state = {
+      showLinks: false
```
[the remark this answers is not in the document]
tobiasbrown: Default value has been changed so all links are visible.

apps/marketplace/components/Brief/Overview/BriefOverviewSectionList.js
tobiasbrown: You're right, this isn't accessible.  Maybe we should use the progress indicator?

apps/marketplace/pages/BriefOverviewPage.js
```diff
@@ -28,23 +29,30 @@ export class BriefOverviewPageComponent extends Component {
 
     return (
       <div>
-        {currentlySending
-          ? <LoadingIndicatorFullPage />
-          : <BriefOverview setFocus={setFocus} {...this.props} briefId={match.params.briefId} />}
+        {currentlySending && <LoadingIndicatorFullPage />}
+        {!currentlySending &&
```
matt-sm: this is confusing use a ternary?

apps/marketplace/components/Brief/Overview/BriefOverview.js
```diff
@@ -1,33 +1,81 @@
...
+    sections: PropTypes.array.isRequired,
+    status: PropTypes.string.isRequired
   }
 
   constructor(props) {
     super(props)
-    this.state = {}
+    this.state = {
+      showDeleteAlert: false,
+      showDeleteButton: false
```
matt-sm: do you need this as internal state?  you can determine whether to display button via `this.props.status`

PR #655, opened 2018-05-07
Uikit upgrade fixes

ndavison:


matt-sm:
Focus not working on most buttons eg. signup, brief response

PR #660, opened 2018-05-16
Feature/new header

marcoalfonso:


apps/marketplace/components/Header/Header.scss
```diff
@@ -1,98 +0,0 @@
-@import "../../main.scss";
```
matt-sm: all the header styles went to custom.css, so they could be shared with other repos?

marcoalfonso: Yes, that way if we change something in there it would change all the headers. It was a pain having to maintain css for three headers. There is also a lot of unused code in the python repos css, but didn't get to clean it up.

src/bundles/Header/AuthWidget.js
```diff
@@ -1,41 +1,17 @@
...
-    loginUrl: PropTypes.string,
-    dashboardUrl: PropTypes.string,
-    logoutUrl: PropTypes.string,
-    isAuthenticated: PropTypes.bool
-}
-
-AuthWidget.defaultProps = {
-    registerText: 'Register',
-    dashboardText: 'Dashboard'
+export const AuthWidget = (props) => {
```
matt-sm: why still bother with this `AuthWidget`?

marcoalfonso: Basically because the of the server rendered isAuthenticated prop. We would have to rebuild the python pages to not serve a server side rendered bundle, and then we would have to consolidate a store that has auth state in the front end apps. Its quite complicated :/

PR #663, opened 2018-05-23
[Domain pricing] Hide VFM message on Pricing page

tobiasbrown:


src/bundles/SellerRegistration/components/PricingForm/PricingForm.js
```diff
@@ -73,12 +73,13 @@ class PricingForm extends BaseForm {
             {Object.keys(validServices).map((service, i) => (
               <fieldset key={`pricing.${service}.${i}`} className="field">
                 <legend>{service}</legend>
-                <label>
+                {/* Hide VFM message until wording is finalised */}
+                {false && <label>
```
matt-sm: i would just delete it - we can always resurrect this code from git if they want it back in

PR #668, opened 2018-05-31
Feature/new opportunities page

marcoalfonso:


apps/marketplace/components/Footer/AUFooter.js
```diff
@@ -34,20 +34,27 @@ const AUFooter = () => (
           <div className="au-footer__end">
             <div className="footer__content footer__legal">
               <p>
-                An initiative of the <a href="https://www.dta.gov.au/">Digital Transformation Agency</a>.{' '}
-                This program forms part of the National Innovation and Science <a href="https://innovation.gov.au/page/agenda">Agenda</a>.{' '}
-                With the exception of the Commonwealth Coat of Arms and where otherwise noted,{' '}
-                this work is licensed under the <a href="https://github.com/govau/uikit/blob/master/LICENSE" rel="external">MIT license</a>
+                An initiative of the <a href="https://www.dta.gov.au/">Digital Transformation Agency</a>. This program
```
matt-sm: assume we dont want these changes to be merged?

marcoalfonso: Yes, that is just prettier linting

apps/marketplace/components/Opportunities/Opportunities.js
```diff
@@ -0,0 +1,310 @@
...
+      </div>
+    )
+  }
+}
+
+Opportunities.propTypes = {}
+
+const mapStateToProps = ({}) => ({})
+
+export default connect(mapStateToProps)(Opportunities)
```
ndavison: Does this need to be a connected component? the `mapStateToProps` is empty so it doesn't seem like it's actually using any redux state?

marcoalfonso: Yes! This is just the UI part, the redux part is missing and that is where this will come in handy. Redux part will be fun :)

ndavison: Ah of course! sorry I forgot it still needs backend integration.

For the person doing the integration: I still suspect the opportunities component won't need to be connected to redux, as the page will do the network call(s) and can pass in the items as a prop.

marcoalfonso: Stateless components are preferred when there is no state management in the component. However this component will require state management because it has so many state changes in the accordion and filtering. Ultimately stateless vs stateful is just a presentation choice, don't affect anything else. Same as composition of dividing a component into subcomponents. Just helps readability. But give time constrains it is all one big blob :P

PR #669, opened 2018-06-01
Mar 1624

ndavison:


src/bundles/SellerRegistration/redux/helpers.js
```diff
@@ -108,4 +108,19 @@ export const focusHeading = () => {
...
+  } else {
+    for(let domain of Object.keys(services)) {
+      if (typeof pricing[domain] === 'undefined') {
+        hasDailyRates = false
+        break
+      }
+    }
+  }
+  return !hasDailyRates
+}
```
tobiasbrown: Would this be easier to follow if you returned at various points?

src/bundles/SellerRegistration/redux/helpers.js
```diff
@@ -108,4 +108,19 @@ export const focusHeading = () => {
...
\ No newline at end of file
+}
+
+export const isDailyRateMissing = (pricing, services) => {
+  let hasDailyRates = true
+  if (!pricing || !services) {
+    hasDailyRates = false
+  } else {
+    for(let domain of Object.keys(services)) {
+      if (typeof pricing[domain] === 'undefined') {
```
tobiasbrown: I think we may need to check the `maxPrice` property.  I had trouble replicating the message due to empty strings.

PR #676, opened 2018-06-15
Feature/new opportunities page

ndavison:


apps/marketplace/actions/opportunitiesActions.js
```diff
@@ -0,0 +1,56 @@
...
+      } else {
+        dispatch(setErrorMessage(`${GENERAL_ERROR} (${response.message})`))
+      }
+    } else {
+      setErrorMessage(GENERAL_ERROR)
+    }
+  }
+}
+
+export const sendingRequest = currentlySending => ({
```
matt-sm: use `appActions.sendingRequest` ?

ndavison: I've found that the app wide `sendingRequest` interferes with this, specifically I think the "ping" request is interfering - i.e. when the ping is finished the loader indicator goes away, but opportunities data is a bit slower and isn't back in time so you get a blank body for a brief period.

A `sendingRequest` specific to opportunities means the timing is on the actual opportunities request.

apps/marketplace/actions/opportunitiesActions.js
```diff
@@ -0,0 +1,56 @@
+import { OPPORTUNITIES_SUCCESS, OPPORTUNITIES_SENDING } from '../constants/constants'
+import { GENERAL_ERROR } from '../constants/messageConstants'
+import dmapi from '../services/apiClient'
+import { setErrorMessage } from './appActions'
+
+export const handleErrorFailure = response => dispatch => {
```
matt-sm: we need to get better at standardizing this - we keep repeating similar

apps/marketplace/components/Opportunities/Opportunities.js
```diff
@@ -0,0 +1,202 @@
...
+    return (
+      <div className={styles.container}>
+        <article className="opportunities-page" role="main">
+          {opportunities.length === 0 && (
+            <div className="row">
+              <div className="col-xs-12">There are no opportunities to show that match your filter.</div>
+            </div>
+          )}
+          {opportunities.length > 0 && (
+            <div className={`${styles.tableDesktop} ${styles.hideMobile}`} role="table" aria-label="Opportunities">
```
matt-sm: do accessibility team have an opinion on the aria tags?

ndavison: I don't think we've heard back on this page yet. I did test using NVDA and I could only get it to announce the table name and column headers (when selecting a cell) using the aria tags, when using div based tables.

apps/marketplace/actions/opportunitiesActions.js
matt-sm: I think the real solution to this is to do the reverse - `ping` needs its own action.  Then others would be free to use `sendingRequest`

ndavison: That's a good point, although in some pages it may make sense not to separate the ping request out (if that page requires data coming from the ping response, e.g. CSRF token, user details).

Perhaps instead the app `sendingRequest` should be an array of inflight requests, that are then removed on completion?

apps/marketplace/components/Opportunities/OpportunitiesFilters.js
```diff
@@ -0,0 +1,465 @@
...
+import AUaccordion from '@gov.au/accordion/lib/js/react.js'
+import { AUcheckbox } from '@gov.au/control-input/lib/js/react.js'
+import AUheading from '@gov.au/headings/lib/js/react.js'
+import styles from './Opportunities.scss'
+
+export class OpportunitiesFilters extends Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      accordionOpen: props.accordionOpen,
```
matt-sm: I realize this is all "internal" state but I still feel like it should be managed in redux

apps/marketplace/components/Opportunities/OpportunitiesFilters.js
```diff
@@ -0,0 +1,465 @@
...
+    }
+    this.handleCheckboxChange = this.handleCheckboxChange.bind(this)
+    this.handleTypeFilterClick = this.handleTypeFilterClick.bind(this)
+    this.handleFilterCancelClick = this.handleFilterCancelClick.bind(this)
+    this.handleFilterApplyClick = this.handleFilterApplyClick.bind(this)
+  }
+
+  setActiveFilterState() {
+    this.setState(curState => {
+      const newState = { ...curState }
```
matt-sm: could this not work like `getFilterValues` ?

ndavison: by referencing `this.state`? I believe when setting state you want to avoid references to `this.state`.

apps/marketplace/components/Opportunities/OpportunitiesFilters.js
matt-sm: I just feel like you've had to do a lot of the boilerplate that we get for free using react-redux-form + redux.  The general pattern we've used in the rest of the components is to use a Form when user submitting.

apps/marketplace/actions/opportunitiesActions.js
matt-sm: That could get tricky to manage - may need to think about this more

tobiasbrown:
What do you think about collapsing the borders on the desktop filters list (not the drop down)?   Does this look a bit better?

**Collapsed**
screen shot 2018-06-19 at 3 50 51 pm

**Not collapsed**
screen shot 2018-06-19 at 4 01 10 pm

apps/shared/ClosedDate.js
```diff
@@ -1,20 +1,59 @@
...
+      } else {
+        final = (
+          <time dateTime={format(parsed, 'YYYY-MM-DDTHH:mm:ss.SSSZ')}>{`6pm, ${format(parsed, 'D MMMM YYYY')}`}</time>
+        )
+      }
     }
   }
 
-  return text
+  return final
```
tobiasbrown: Just wondering whether using something like [moment.js](https://momentjs.com/) would simplify and add flexibility if we need to do more date calculations? What do you think @matt-sm?

public/custom.css
```diff
@@ -297,13 +297,83 @@ label.text-box-number-label {
...
 
-.au-accordion {
+/* .au-accordion {
   margin-bottom: -1px !important;
 }
 
 .au-body .au-accordion__title, .au-accordion__title {
   font-size: 17px !important;
   padding: 16px 14px !important;
+} */
```
tobiasbrown: Is this a WIP or can we remove completely?

public/custom.css
```diff
@@ -297,13 +297,83 @@ label.text-box-number-label {
...
+
+.opportunities-page .au-accordion {
+  margin-top: -4px;
+  border: none;
+}
+
+.opportunities-page .au-accordion__title {
+  background: white;
+  padding: 6px 16px 9px 16px !important;
```
tobiasbrown: I think you can remove `!important`.

public/custom.css
```diff
@@ -297,13 +297,83 @@ label.text-box-number-label {
...
+  border: solid 3px #e0e0e0;
+  border-radius: 4px;
+  color: #313131;
+  margin: 0 !important;
```
tobiasbrown: I think you can remove `!important`.

apps/shared/ClosedDate.js
matt-sm: yeah - possibly - but I am very hesitant to add another dependency

public/custom.css
```diff
@@ -297,13 +297,74 @@ label.text-box-number-label {
...
+}
+
+.opportunities-page .au-accordion__body {
+  overflow: visible;
+}
+
+.opportunities-page .au-accordion__title:focus, .opportunities-page .au-accordion__title:hover {
+  outline: solid 3px #45c2f0;
+  outline-offset: -3px;
+  color: #00698f !important;
```
tobiasbrown: You can remove this `!important`.

PR #683, opened 2018-07-12
Mar 1668

ndavison:


apps/marketplace/components/BriefChoice/BriefChoice.js
```diff
@@ -1,79 +1,85 @@
...
+    <div className={`row ${styles.headingRow}`}>
+      <div className="col-xs-12">
+        <AUheading level="1" size="xl">
+          Create a new brief
+        </AUheading>
+      </div>
+    </div>
+    <div className={`row ${styles.flex}`}>
+      <div className={`col-xs-12 col-md-3 ${styles.flexColumn}`}>
+        <span />
```
tobiasbrown: Can we remove these empty `<span />` tags?  Do they have a purpose?

apps/marketplace/components/BriefChoice/BriefChoice.js
```diff
@@ -1,79 +1,85 @@
...
+        <AUheading level="2" size="md">
+          Digital outcome
+        </AUheading>
+        <p>Find a team to help you work towards a digital project or goal, for example, a website.</p>
+        <ul className={styles.tickList}>
+          <li>Receive written proposals from sellers.</li>
+          <li>Request case studies, work history, references and/or presentations.</li>
+        </ul>
+        <p className={styles.flexGrow}>
+          <strong>Receive responses within:</strong>
```
tobiasbrown: Does this need to be in a `<strong>` tag?  Maybe just style with `font-weight`?

apps/marketplace/components/BriefChoice/BriefChoice.js
```diff
@@ -1,79 +1,85 @@
...
+          <br />1 - 2 weeks
```
tobiasbrown: `1 - 2` won't meet accessibility guidelines.  We need to update this to `1 to 2` or something more appropriate.

apps/marketplace/components/BriefChoice/BriefChoice.js
ndavison: Because the `1 - 2 weeks` part isn't bold, this would require adding some tag around the text anyway e.g. `span` to add the class for the font-weight. Using `strong` also means it will use default uikit bold text styling.

PR #685, opened 2018-07-13
Changes to the admin "show detailed changes" table

ndavison:
This PR will: add pricing and recruiter changes to the admin detailed changes view, and relax some of the requirements so more changes are displayed, including deeper level object changes (represented as a stringified object).


src/bundles/ApplicationsAdmin/components/AppChangesTable/AppChangesTable.js
```diff
@@ -39,6 +39,9 @@ export const AppChangesTable = props => {
                       (typeof change.value === 'boolean' ? displayBoolean(change.value) :
                         null)
                   )}
+                  {typeof change.value === 'object' && (
```
matt-sm: not sure about this

ndavison: This is sort of a catch all, to at least show a stringified version of the change. For something like a seller adding a price to a newly added domain, due to the way the diffs are done we don't get the max price but only the fact they added a new domain to the pricing array - this will show that but the value will also show the max price.

We do the diffs using a library with limited config options, and changing it is going to be a pretty big task.

PR #690, opened 2018-07-24
#MAR-1805 Add training automation and take into account the new header

malmas18:


fixtures/flows/utils.js
```diff
@@ -55,10 +55,19 @@ export const clickButton = async (value) => {
     await button.click();
 }
 
-export const clickLink = async (linkText) => {
+export const clickLink = async (linkText, isUrl) => {
     console.log(`Clicking link "${linkText}"`);
-    let links = await getElementHandles(`//a[.="${linkText}"]`);
-    expect(links.length).to.equal(1, `Number of links found for "${linkText}"=${links.length}`);
+    var links;
```
tobiasbrown: Can you use `let` instead of `var`?

malmas18: no. `let` is undefined in `if`

PR #714, opened 2018-08-28
Mar 2376 seller search component

ndavison:
This is an initial stab at a standalone React component for a supplier search input, with auto complete suggestions for results that match the current query.

This has not yet been tested for a11y.

### DIFF
--- a/apps/marketplace/actions/supplierActions.js
+++ b/apps/marketplace/actions/supplierActions.js
@@ -1,0 +1,16 @@
+import dmapi from '../services/apiClient'
+
+const findSuppliers = keyword => {
+  const params = {
+    keyword
+  }
+  return dmapi({ url: `/suppliers/search`, params }).then(response => {
+    if (!response || response.error) {
+      throw response.errorMessage
+    } else {
+      return response.data
+    }
+  })
+}
+
+export default findSuppliers
--- a/apps/marketplace/components/SellerSelect/SellerSelect.js
+++ b/apps/marketplace/components/SellerSelect/SellerSelect.js
@@ -43,11 +43,7 @@ export class SellerSelect extends Component {
     this.setState(curState => {
       const newState = { ...curState }
       newState.sellers = []
-      if (this.props.showSelected) {
-        newState.inputValue = seller.name
-      } else {
-        newState.inputValue = ''
-      }
+      newState.inputValue = this.props.showSelected ? seller.name : ''
       this.props.onSellerSelect(seller)
       return newState
     })
@@ -66,7 +62,7 @@ export class SellerSelect extends Component {
           placeholder={this.props.placeholder}
           onChange={this.handleOnChange}
           value={this.state.inputValue}
-          className={this.props.showSearchButton ? styles.noRightCorners : ''}
+          className={this.props.showSearchButton ? styles.noRightRadius : ''}
         />
         {this.props.showSearchButton && <AUbutton onClick={this.handleSearchClick}>Search</AUbutton>}
         {this.state.sellers && (
@@ -76,11 +72,8 @@ export class SellerSelect extends Component {
                 <a href={`#${seller.code}`} onClick={e => this.handleSellerSelectClick(seller, e)}>
                   {seller.name}
                 </a>
-                {seller.sme && (
-                  <span className={styles.smeBadge}>
-                    <abbr title="Small medium enterprise">SME</abbr>
-                  </span>
-                )}
+                {seller.panel && <span className={`${styles.badge} ${styles.panelBadge}`}>✓ Panel</span>}
+                {seller.sme && <span className={`${styles.badge} ${styles.smeBadge}`}>SME</span>}
               </li>
             ))}
           </ul>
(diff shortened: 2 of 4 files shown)


apps/marketplace/actions/supplierActions.js
```diff
@@ -0,0 +1,16 @@
...
+  return dmapi({ url: `/suppliers/search`, params }).then(response => {
+    if (!response || response.error) {
+      throw response.errorMessage
+    } else {
+      return response.data
+    }
+  })
+}
+
+export default findSuppliers
```
ndavison: Needed to make this default as eslint prefers a file with only one export to use `default` apparently.

apps/marketplace/actions/supplierActions.js
```diff
@@ -0,0 +1,16 @@
+import dmapi from '../services/apiClient'
+
+const findSuppliers = keyword => {
+  const params = {
+    keyword
+  }
+  return dmapi({ url: `/suppliers/search`, params }).then(response => {
+    if (!response || response.error) {
+      throw response.errorMessage
```
ndavison: The use of `throw` here is a bit different to how we've previously been using `dmapi()`. A goal of this component was to make it without a dependency on React Redux, so the whole `dispatch()` system is unavailable. It turns out that `dmapi()` and its response object is a `Promise`, which means handling success and fail can be done without callbacks - this is pointed out in the component code below.

apps/marketplace/components/SellerSelect/SellerSelect.js
```diff
@@ -0,0 +1,104 @@
...
+    })
+
+    if (keyword && keyword.length > 2) {
+      findSuppliers(keyword)
+        .then(data => {
+          this.setState({
+            sellers: data.sellers
+          })
+        })
+        .catch(() => {})
```
ndavison: This catches the `throw` condition in the `dmapi()` usage in the `supplierActions.js` code. Currently, it's not doing anything with error responses from the API.

apps/marketplace/components/SellerSelect/SellerSelect.js
```diff
@@ -0,0 +1,104 @@
...
+
+  handleOnChange(e) {
+    const keyword = e.target.value
+    this.setState({
+      inputValue: keyword
+    })
+
+    if (keyword && keyword.length > 2) {
+      findSuppliers(keyword)
+        .then(data => {
```
ndavison: `dmapi()` and its response returns a `Promise` so we can use `then()` etc directly in the component rather than use redux's dispatch system to handle responses and state changes.

apps/marketplace/components/SellerSelect/SellerSelect.js
```diff
@@ -0,0 +1,126 @@
...
+      placeholder={props.placeholder}
+      onChange={props.handleOnChange}
+      value={props.inputValue}
+      className={props.className}
+    />
+    {props.showSearchButton && <AUbutton onClick={props.handleSearchClick}>Search</AUbutton>}
+  </div>
+)
+
+const SellerSelectResultsView = props => (
```
ndavison: In this instance separating the components into view/UI and data components is probably not useful but I think it's a pattern that could be very useful going forward for components that could be used in multiple places.

apps/marketplace/components/SellerSelect/SellerSelect.js
```diff
@@ -0,0 +1,134 @@
...
+    const keyword = e.target.value
+    this.setState({
+      inputValue: keyword
+    })
+
+    if (timeoutHandle) {
+      clearTimeout(timeoutHandle)
+    }
+
+    timeoutHandle = setTimeout(() => {
```
ndavison: the `setTimeout()` is to prevent flooding the network with requests after the 2 chars minimum. When the `onChange` is fired for the input, it will cancel any unprocessed `setTimeout` currently pending, preventing previous key strokes from impacting on the UI results should network requests not come back in order.

apps/marketplace/components/SellerSelect/SellerSelect.js
```diff
@@ -0,0 +1,134 @@
...
+)
+
+const SellerSelectResultsView = props => (
+  <ul className={props.className}>
+    {props.sellers.map(seller => (
+      <li key={seller.code}>
+        <a href={`#${seller.code}`} onClick={e => props.handleSellerSelectClick(seller, e)}>
+          {seller.name}
+        </a>
+        {seller.panel && <span className={props.panelBadgeClassName}>✓ Panel</span>}
```
tobiasbrown: The check mark may need to be in a separate `<span>` so that screen readers ignore it.  See https://www.w3.org/WAI/GL/wiki/Unicode_Character_with_an_On-Screen_Text_Alternative.  Might be a question for a11y team?

apps/marketplace/components/SellerSelect/SellerSelect.js
```diff
@@ -0,0 +1,134 @@
...
+
+    timeoutHandle = setTimeout(() => {
+      if (keyword && keyword.length > 2) {
+        findSuppliers(keyword)
+          .then(data => {
+            this.setState({
+              sellers: data.sellers
+            })
+          })
+          .catch(() => {})
```
tobiasbrown: Just adding a note here to review error handling when this component is used for the first time.

PR #719, opened 2018-09-17
[Application] Allow expiry date edits when new documents are uploaded

tobiasbrown:
This pull request:
* prevents expiry dates from being edited without a new document being uploaded
* resets documents when the expiry date has passed or is the current day

Addresses MAR-2398


src/bundles/SellerRegistration/components/DocumentsForm/DocumentsForm.js
```diff
@@ -38,8 +40,17 @@ class DocumentsForm extends BaseForm {
     static defaultProps = {
         match: {url: ''}
     }
+    
     state = {
-        errors: {}
+        errors: {},
+        liability: {
```
malmas18: `uploading` and `file` should be initialised here?

src/bundles/SellerRegistration/components/DocumentsForm/DocumentsForm.js
```diff
@@ -38,8 +40,17 @@ class DocumentsForm extends BaseForm {
...
+            newDocumentUploaded: this.props.documentsForm.documents.liability && 
+                this.props.documentsForm.documents.liability.expiry ? false : true
+        },
+        workers: {
+            newDocumentUploaded: this.props.documentsForm.documents.workers &&
```
malmas18: `uploading` and `file` should be initialised here?

PR #725, opened 2018-10-02
Add the CSRF token into API requests that use an unsafe HTTP method

ndavison:
With https://github.com/AusDTO/dto-digitalmarketplace-api/pull/534 enabling CSRF token handling on `/api/2` endpoints, we need to make the frontend send the CSRF token when it is making an unsafe HTTP method request. This PR modifies the usages of the `dmapi` service so it includes the token in the header of such requests.

The token itself is already included in the app state as part of the `/api/2/ping` request made on initial page load.


apps/marketplace/actions/appActions.js
```diff
@@ -72,12 +72,16 @@ export const fetchAuth = () => dispatch => {
...
 
-export const login = data => dispatch => {
+export const login = data => (dispatch, getState) => {
   dispatch(sendingRequest(true))
   dispatch(setAuthFrameworkError(false))
   dmapi({
     method: 'post',
     url: '/login',
+    headers: {
+      'X-CSRFToken': getState().app.csrfToken,
```
ndavison: There is some talk in the React community that using `getState` in a redux action is a bad pattern, breaking the 1 way flow of data, among other concerns. On the other hand, [some wisdom suggests it serves a purpose](https://blog.isquaredsoftware.com/2017/01/idiomatic-redux-thoughts-on-thunks-sagas-abstraction-and-reusability/) and I think in this case adding a global state value like a CSRF token to each action that needs it makes sense. The alternative approach was to add the token as a hidden field in the render of all relevant components/pages and is still possible if there is strong objections to the `getState` approach.

PR #730, opened 2018-11-01
Redirect Collaborate

ndavison:
This PR removes the Collaborate page and components, and redirects to the infrastructure.gov.au page.


apps/marketplace/routes.js
```diff
@@ -27,9 +26,13 @@ export const Routes = () => (
     <PrivateRoute path={`${rootPath}/brief/:briefId`} component={BriefPage} />
     <Route path={`${rootPath}/reset-password`} component={ResetPasswordPage} />
     <Route path={`${rootPath}/login`} component={LoginPage} />
-    <Route path={`${rootPath}/collaborate/funded/:state`} component={CollaboratePage} />
-    <Route path={`${rootPath}/collaborate/:stage`} component={CollaboratePage} />
-    <Route path={`${rootPath}/collaborate`} component={CollaboratePage} />
+    <Route
+      path={`${rootPath}/collaborate`}
+      component={() => {
+        window.location = 'https://infrastructure.gov.au/cities/smart-cities/collaboration-platform/'
```
ndavison: We must do it this way as the `Redirect` react router component doesn't do external redirects.

malmas18: the collaborate lines should be deleted.

add to `server/render.js`

```
app.get('/2/collaborate*', function(req, res) {
  res.redirect('https://infrastructure.gov.au/cities/smart-cities/collaboration-platform/');
})
```

PR #738, opened 2019-01-25
[MAR-2616] add atm filter

malmas18:


apps/marketplace/components/Opportunities/OpportunitiesFilters.js
```diff
@@ -62,17 +63,33 @@ export class OpportunitiesFiltersComponent extends BaseForm {
...
+          statusAccordionOpen: false,
+          typeAccordionOpen: false
+        })
+        break
+      case 'type':
+        this.setState({
+          typeAccordionOpen: isOpen,
+          locationAccordionOpen: false,
+          statusAccordionOpen: false
+        })
```
tobiasbrown: Can we reduce some repetition in this `switch` if we did something like this:
```
case 'location':
case 'status':
case 'type:
  this.setState({
      locationAccordionOpen: type == 'location' ? isOpen : false,
      statusAccordionOpen: type == 'status' ? isOpen : false,
      typeAccordionOpen: type == 'type' ? isOpen : false
  })
```

malmas18: done

apps/marketplace/components/Opportunities/OpportunitiesFilters.js
```diff
@@ -8,7 +8,6 @@ import { changeForm, loadOpportunities } from 'marketplace/actions/opportunities
 import CheckboxDetailsField from 'shared/form/CheckboxDetailsField'
 import AUaccordion from '@gov.au/accordion/lib/js/react.js'
 import AUheading from '@gov.au/headings/lib/js/react.js'
-import TypeFilterControl from './TypeFilterControl'
```
ndavison: You should be able to delete the `TypeFilerControl.js` file and the `a.filter` and `a:hover.filter, a.filter.active` rules in `Opportunities.scss`.

ndavison: And the `li.filterContainer` rule.

PR #743, opened 2019-02-25
Mar 2897 fix signup process

ndavison:


apps/marketplace/actions/memberActions.js
```diff
@@ -42,50 +40,59 @@ export const handleSignupSubmit = model => (dispatch, getState) => {
...
+    headers: {
+      'X-CSRFToken': getState().app.csrfToken,
+      'Content-Type': 'application/json'
+    },
+    data: { password }
+  }).then(response => {
     if (response.error) {
-      dispatch(setErrorMessage(REGISTRATION_NOT_FOUND))
+      if (response.status === 409) {
+        dispatch(setErrorMessage(DUPLICATE_USER))
```
malmas18: Doesn't this leak email addresses if tried enough times?

I also thought that the system would send an email about forgotten password when an email match happens.

apps/marketplace/pages/ResetPasswordPage.js
```diff
@@ -43,10 +36,27 @@ export class ResetPasswordPageComponent extends Component {
...
+  }
+
+  getEmailFromQueryString() {
+    const parsed = parse(this.props.location.search.substr(1))
+    let emailAddress = ''
+    if (parsed.e) {
+      emailAddress = parsed.e
+    }
+    return emailAddress
+  }
```
tobiasbrown: Should we move these two functions, `getTokenFromURL ` and `getEmailFromQueryString `, into a helper file since they're also used by the `CreateUserPage` component?

apps/marketplace/actions/memberActions.js
ndavison: This is part of the create user request, which should now only process with a valid token and email address pair, which should not be created if an email address is already linked to an account - it will do as you said, i.e. say the signup was successful and to look out for an email, but the email sent is the one saying you're already a user. So I don't think this duplicate user check is relevant anymore.

The API will only respond with a 409 if a seller tries to signup using a (non generic) email domain that is already associated with an existing supplier (in which case an existing account needs to use the 'invite contributor' feature).

PR #759, opened 2019-04-02
Validation of seller profile URLs

ndavison:
Add extra validation to ensure the URLs provided by sellers in their application are properly formed.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,4 @@
 **/*.js
 !apps/
+!fixtures/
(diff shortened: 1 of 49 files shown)


ndavison:
> Do we need the same validation on the back end?

There is already similar validation that these values are links, however they can be bypassed, which is what https://github.com/AusDTO/dto-digitalmarketplace-api/pull/615 is for - to ensure the application can't be submitted if the validations failed.

PR #785, opened 2019-06-21
[MAR-3300] fix missing signed agreement

malmas18:


src/bundles/SellerRegistration/redux/modules/reducer.js
```diff
@@ -9,16 +9,22 @@ import signup from './signup';
...
 import errorMessage from '../../../../shared/reduxModules/errorMessage';
 import form_options from '../../../../shared/reduxModules/form_options';
+import application_errors from './application_errors';
+import agreement from './agreement'
+import domains from './domains'
 
 export default combineReducers({
+  agreement,
   application,
+  application_errors,
```
tobiasbrown: Should we use camel case?

malmas18: annoyingly, this needs to match the response

src/shared/SellerProfile/Body.js
```diff
@@ -279,11 +279,15 @@ const Body = (props) => {
...
-          (<span><a href={signed_agreements[0]['agreement']['url']}>{signed_agreements[0]['agreement']['version']}
-          </a> signed on {signed_agreements[0]['signed_at']
-            && format(new Date(signed_agreements[0]['signed_at']), 'DD/MM/YYYY')}</span>)
-          }
+          {signed_agreements && signed_agreements.map((sa, i) => (
+            <React.Fragment key={i}>
+              {sa['agreement'] && <div>
+                <a href={sa['agreement']['url']}>{sa['agreement']['version']}</a>{' '}
+                signed on {sa['agreement']['signed_at'] && format(new Date(sa['agreement']['signed_at']), 'DD/MM/YYYY')}
+                {sa['signed_at'] && format(new Date(sa['signed_at']), 'DD/MM/YYYY')}
```
ndavison: Are there supposed to be two dates being output here? i.e. what happens if `sa['agreement']['signed_at']` and `sa['signed_at']` are both truthy?

malmas18: updated to only ever show one

PR #808, opened 2019-08-13
Expand recruiter badge in admin portal

ndavison:
When viewing a full supplier profile in the admin portal, ops has requested that the 'Recruiter' badge be expanded to be more specific as to the exact recruiter status of the supplier. This PR changes the badge to say `Recruiter`, `Consultant` or `Recruiter and Consultant` depending on the status.


src/shared/Badges/Badges.css
```diff
@@ -23,6 +23,15 @@
 .badge__recruitment {
   background-color: #880e48;
 }
+.badge__recruiter_only {
+  background-color: #880e48;
+}
+.badge__recruiter_both {
+  background-color: #880e48;
+}
+.badge__consultant_only {
```
tobiasbrown: Do we need these classes if the colours are the same?  Should we use `badge__recruitment`?

ndavison: It's because of the `titleMap` object in `Badges.js` - the component is using that to create the different badges and also uses the same value as the class. I figured it was cleaner to just add more classes than change the component to split out the badge's text key and class.

PR #811, opened 2019-08-15
[MAR-3415] update validation

malmas18:


apps/marketplace/components/SellerAssessment/SellerAssessmentCriteriaStage.js
```diff
@@ -83,21 +104,13 @@ class SellerAssessmentCriteriaStage extends Component {
           title="An error occurred"
           model={this.props.model}
           messages={{
-            requiredMinimal: `You must select at least ${getCriteriaNeeded(
-              domain.criteriaNeeded,
-              domain.priceMaximum,
-              this.props[this.props.model].maxDailyRate
-            )} criteria`
+            requiredMinimal: `You must submit evidence for at least ${criteriaNeeded} criteria`,
```
tobiasbrown: I think we should cater for the scenario when only one criterion is required.  For example: `You must submit evidence for at least 1 criterion`.

PR #814, opened 2019-08-20
Remove the onChange from the shared Checkbox component

ndavison:
This `onChange()` property on the shared checkbox is not being used anywhere in code, and complicates its API, so this PR is removing it.

### DIFF
--- a/apps/shared/form/CheckboxDetailsField.js
+++ b/apps/shared/form/CheckboxDetailsField.js
@@ -30,18 +30,14 @@ class CheckboxDetailsField extends React.Component {
   }
 
   render() {
-    const { name, id, label, model, validators, messages, onChange, value, description, disabled } = this.props
+    const { name, id, label, model, validators, messages, value, description, disabled } = this.props
     /* eslint-disable jsx-a11y/label-has-for */
 
     return (
       <label className="au-control-input" htmlFor={id}>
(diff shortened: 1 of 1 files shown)


apps/shared/form/CheckboxDetailsField.js
```diff
@@ -30,18 +30,14 @@ class CheckboxDetailsField extends React.Component {
...
     return (
       <label className="au-control-input" htmlFor={id}>
         <Control.checkbox
           className="au-control-input__input"
-          onClick={e => {
-            this.onToggle.bind(this)
-            this.onToggle(e)
-            onChange(e)
-          }}
+          onClick={this.onToggle.bind(this)}
```
malmas18: does this actually call onToggle now with the correct context?
onToggle uses `this` and `e`.

I would think to add `this.onToggle.bind(this)` in the constructor and call `this.onToggle(e)` from here.

```suggestion
          onClick={e => this.onToggle(e)}
```

ndavison: This is how it was originally (before the update dependencies PR) https://github.com/AusDTO/dto-digitalmarketplace-frontend/blame/633cabc61de927153478e26d22f57326f147ca5e/apps/shared/form/CheckboxDetailsField.js#L40

PR #824, opened 2019-08-29
[Opportunity] Show values if they've been set

tobiasbrown:
This pull request will only show values for these fields if they've been set.

* Questions close date
* Closing date
* Published date
* Start date
* Budget range

Addresses MAR-3387


apps/marketplace/components/Opportunity/Opportunity.js
```diff
@@ -251,7 +260,7 @@ const Opportunity = props => {
                 <div className="col-xs-12 col-sm-4">
                   <strong>Budget range</strong>
                 </div>
-                <div className="col-xs-12 col-sm-8">{brief.budgetRange || 'None specified'}</div>
+                <div className="col-xs-12 col-sm-8">{brief.budgetRange && brief.budgetRange}</div>
```
ndavison: This will be an empty div if there is no `budgetRange`, but the "Budget range" row itself will still show. Should we use a `{brief.budgetRange && (...` which wraps around this entire row? or add the check to the existing `brief.lotSlug === 'rfx'` above.

tobiasbrown: I've added the check up higher so it hides the row (like it does for specialist opportunities).

apps/marketplace/components/Opportunity/Opportunity.js
```diff
@@ -216,7 +223,9 @@ const Opportunity = props => {
                 <strong>Estimated start date</strong>
               </div>
               <div className="col-xs-12 col-sm-8">
-                {brief.lotSlug === 'specialist' ? format(brief.startDate, 'D-MM-YYYY') : brief.startDate}
+                {brief.lotSlug === 'specialist' && brief.startDate
```
malmas18: doesn't this still display it?

apps/marketplace/components/Opportunity/Opportunity.js
```diff
@@ -74,6 +74,18 @@ const getBriefCategory = (domains, briefCategory) => {
...
+const getQuestionsCloseDate = brief => {
+  if (brief.dates.questions_close) {
+    return new Date(brief.dates.questions_close)
+  }
+
+  if (getClosingTime(brief)) {
+    return getBriefLastQuestionDate(new Date(getClosingTime(brief)))
+  }
+
+  return false
```
malmas18: keep return types consistent..
```suggestion
  return null
```

apps/marketplace/components/Opportunity/Opportunity.js
tobiasbrown: This is to prevent `Invalid Date` from showing.

Screen Shot 2019-09-05 at 9 49 40 am

apps/marketplace/components/Opportunity/Opportunity.js
tobiasbrown: Should we return `null` for `showATMObjectives`?

apps/marketplace/components/Opportunity/Opportunity.js
tobiasbrown: I think the default values for non-specialist were ok but I've moved this logic into a function.

PR #828, opened 2019-09-03
[MAR-3308] Download reports

anthonymckerron:
Build reports download page.
Includes:
- Training2 Opportunities
- RFx Opportunities
- ATM Opportunities
- Specialist Opportunities
- All seller responses (bar old training flow)
- Current approved seller catalogue.


apps/marketplace/actions/downloadReportsActions.js
```diff
@@ -0,0 +1,14 @@
+import dmapi from '../services/apiClient'
+
+const submitDownloadReportsRequest = values => (dispatch, getState) =>
```
malmas18: is this actually used?

anthonymckerron: It would seem not. I will remove it.

apps/marketplace/components/validators.js
```diff
@@ -59,6 +63,16 @@ export const dateIs2DaysInFuture = val => {
   return false
 }
 
+export const startDateIsBeforeEndDate = (start, end) => {
+  // if (!validDate(start) || !validDate(end)) {
+  //   return false
+  // }
```
tobiasbrown: What's the plan with this?  Delete?

apps/marketplace/pages/DownloadReports/DownloadReports.scss
```diff
@@ -0,0 +1,9 @@
+.inputsContainer {
+    display: flex;
+    margin-top: 0.75rem;
+  }
+  
+.inputsContainer div {
```
tobiasbrown: Should we nest the `div` styles in the above block?

apps/marketplace/components/validators.js
```diff
@@ -35,15 +35,19 @@ export const requiredFile = val => {
...
   }
-  if (isFuture(val)) {
+  if (enableFutureCheck === undefined || enableFutureCheck) {
+    if (isFuture(val)) {
+      return true
+    }
+  } else {
     return true
   }
   return false
```
tobiasbrown: Is this last `return` redundant with the last `if` block?

apps/marketplace/components/validators.js
```diff
@@ -35,15 +35,19 @@ export const requiredFile = val => {
...
     return false
   }
   if (format(dateObj, 'YYYY-MM-DD') !== val) {
     return false
   }
-  if (isFuture(val)) {
+  if (enableFutureCheck === undefined || enableFutureCheck) {
+    if (isFuture(val)) {
+      return true
+    }
```
tobiasbrown: This might need to return false.
```javascript
return isFuture(val) ? true : false

File: apps/marketplace/pages/DownloadReports/DownloadReports.js [hunk 6]
```diff
@@ -0,0 +1,213 @@
...
+import ErrorAlert from 'marketplace/components/Alerts/ErrorAlert'
+import { Form, actions } from 'react-redux-form'
+import { AUcheckbox } from '@gov.au/control-input'
+import { Redirect } from 'react-router-dom'
+import { connect } from 'react-redux'
+import { hasPermission } from 'marketplace/components/helpers'
+import { rootPath } from 'marketplace/routes'
+import { required, validDate, startDateIsBeforeEndDate } from 'marketplace/components/validators'
+
+import styles from '../../main.scss'
```
tobiasbrown: Should these styles be in `DownloadReports.scss`?

apps/marketplace/pages/DownloadReports/DownloadReports.js
```diff
@@ -0,0 +1,213 @@
...
+              })
+            }}
+            options={[
+              {
+                value: 'sellersCatalogue',
+                label: 'Current approved sellers catalogue'
+              },
+              {
+                value: 'sellerResponses',
+                label: 'Seller responses to your briefs'
```
tobiasbrown:
```suggestion
                label: 'Seller responses to your opportunities'
```

apps/marketplace/pages/DownloadReports/DownloadReports.js
```diff
@@ -0,0 +1,213 @@
...
+                    this.setState({
+                      acceptEnabled: e.target.checked
+                    })
+                  }}
+                  label="I am authorised to access procurement reporting documents across my organisation and I will store, transmit and use the report in line with my agency's privacy and data policies."
+                />
+                <AUbutton type="submit" disabled={!this.state.acceptEnabled}>
+                  Download reports
+                </AUbutton>
+              </div>
```
tobiasbrown: Why are there two submit buttons and checkboxes?

apps/marketplace/components/validators.js
malmas18: to return true if not valid.  valid check should be done elsewhere

apps/marketplace/pages/DownloadReports/DownloadReports.scss
malmas18: file deleted

apps/marketplace/components/validators.js
malmas18: it will be redundant to the default return at the end of the function

apps/marketplace/pages/DownloadReports/DownloadReports.js
malmas18: reusing common styles for spacing

apps/marketplace/pages/DownloadReports/DownloadReports.js
malmas18: removed

apps/marketplace/components/validators.js
malmas18: return false as default

PR #833, opened 2019-09-11
[Textfield] Count remaining characters

tobiasbrown:
This pull request replaces the `[number] characters maximum` message with a count down of how many characters can be entered.

Addresses MAR-3387

### Before
Screen Shot 2019-09-11 at 4 07 41 pm

### After
Screen Shot 2019-09-11 at 4 15 52 pm


tobiasbrown:
> this looks fine, I am just wondering if it is worth while to show it by default instead?
> 
> it looks like where this is configured to show is basically everywhere.

I've removed the prop and set it to display by default.

apps/shared/form/Textfield/Textfield.js
```diff
@@ -0,0 +1,92 @@
...
+    pattern,
+    placeholder,
+    readOnly,
+    showCharacterCounter,
+    type,
+    value
+  } = props
+
+  return (
+    <div>
```
malmas18:
```suggestion
    <React.Fragment>
```

apps/shared/form/Textfield/Textfield.js
```diff
@@ -0,0 +1,92 @@
...
+        name={name}
+        onChange={onChange}
+        pattern={pattern}
+        placeholder={placeholder}
+        readOnly={readOnly}
+        type={type}
+        value={value}
+      />
+      {showCharacterCounter && maxLength > 0 && <CharacterCounter limit={maxLength} value={value} />}
+    </div>
```
malmas18:
```suggestion
    </React.Fragment>
```

PR #852, opened 2019-10-30
[MAR-3537] Add agency edit

malmas18:


src/bundles/ApplicationsAdmin/components/AgencyList/AgencyList.js
```diff
@@ -0,0 +1,92 @@
...
+      loading: true
+    })
+    this.props.loadAgencies()
+      .then(r => {
+        if (r.status === 200) {
+          return r.json()
+        }
+        this.setState({
+          loading: false
+        })
```
ndavison: Is this necessary? it looks like the `loading: false` will be set in the next `.then()` anyway.

src/bundles/ApplicationsAdmin/components/AgencyList/AgencyList.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  componentDidMount() {
+    this.setState({
+      loading: true
+    })
+    this.props.loadAgencies()
+      .then(r => {
+        if (r.status === 200) {
+          return r.json()
```
ndavison: Should this return `Promise.resolve(r.json())`? Or perhaps just set the `agencies` state here and avoid having the extra `.then()`?

src/bundles/ApplicationsAdmin/components/AgencyList/AgencyList.js
```diff
@@ -0,0 +1,92 @@
...
+        }
+        this.setState({
+          loading: false
+        })
+        return Promise.reject()
+      })
+      .then(r => {
+        this.setState({
+          loading: false,
+          agencies: r
```
ndavison: If the promise is rejected above, then `r` here will not likely be in an expected format. If the `agencies` state change is moved to the previous `.then()` inside the success condition, then this `.then()` can be removed.

src/bundles/ApplicationsAdmin/components/Agency/Agency.js
```diff
@@ -0,0 +1,194 @@
...
+      this.setState({
+        loading: false
+      })
+      return Promise.reject()
+    })
+    .then(r => {
+      this.setState({
+        loading: false,
+        agency: r,
+        saved: new Date()
```
ndavison: These state changes can probably occur inside the `r.status === 200` condition in the `.then()` above and this `.then()` can be removed, as right now I think these will be called if the Promise is rejected.

src/bundles/ApplicationsAdmin/components/Agency/Agency.js
```diff
@@ -0,0 +1,194 @@
...
+      data.reports = true
+    } else {
+      data.reports = false
+    }
+    if (data.whitelisted === "on") {
+      data.whitelisted = true
+    } else {
+      data.whitelisted = false
+    }
+     data.domains = data.domains.split('\n')
```
ndavison:
```suggestion
    data.domains = data.domains.split('\n')
```

src/bundles/ApplicationsAdmin/components/AgencyList/AgencyList.js
malmas18: r.json() returns a promise...

src/bundles/ApplicationsAdmin/components/Agency/Agency.js
malmas18: updated

PR #860, opened 2019-11-12
[MAR-3416] Content update

malmas18:


apps/marketplace/components/BuyerBriefFlow/Overview.js
```diff
@@ -130,8 +130,24 @@ class Overview extends Component {
           : 0
 
       const briefResponseCount = briefResponses && briefResponses.length > 0 ? briefResponses.length : 0
-      let flowName = flow !== 'specialist' ? flow.toUpperCase() : 'specialist'
-      flowName = flow === 'training2' ? 'training' : flowName
+      let flowName
```
tobiasbrown: Just wondering if this should be moved to a helper function?  Would we need to do this in other places?

malmas18: used an existing function in helper

PR #865, opened 2019-11-15
[MAR-3103] update content for opp tab in seller dashboard

malmas18:


apps/marketplace/components/SellerDashboard/Opportunities.js
```diff
@@ -100,7 +100,12 @@ export class Opportunities extends Component {
               </tbody>
             </table>
           ) : (
-            <div className={`${styles.marginTop3} ${styles.marginBottom3}`}>Apply for opportunities</div>
+            <div className={`${styles.marginTop1} ${styles.marginBottom3} ${styles.marginLeft1}`}>
+              <p>Your submitted applications will appear here.</p>
```
tobiasbrown: I think the content could be improved to something like this.  What do you think?
```suggestion
              <p>Your responses to opportunities will appear here.</p>
```

malmas18: updated

PR #889, opened 2020-02-03
[Errors] Prevent buyers from accessing pages if opportunity is withdrawn

tobiasbrown:
Addresses MAR-3756


apps/marketplace/pages/BriefPage.js
```diff
@@ -57,27 +58,62 @@ class BriefPage extends Component {
             <Route
               path={`${match.url}/published`}
               render={() => (
-                <BriefSubmitted
-                  setFocus={setFocus}
-                  submitClicked={this.state.submitClicked}
-                  handleSubmit={values => this.handleFeedbackSubmit(values)}
-                  {...this.props}
-                />
+                <span>
```
malmas18: React.Fragment?

tobiasbrown: I've replaced these spans with fragments.

PR #898, opened 2020-02-10
Ask for ABN on signup when creating a supplier account

ndavison:
This changes the signup form to include an ABN field for when a seller signs up. It also changes the seller profile form to always disable the ABN field (as this value is only collected and validated on signup).

This coincides with https://github.com/AusDTO/dto-digitalmarketplace-api/pull/733.


src/bundles/SellerRegistration/components/BusinessDetailsForm/BusinessDetailsForm.js
```diff
@@ -126,16 +127,14 @@ class BusinessDetailsForm extends BaseForm {
...
-                              (<span>You need an Australian Business Number to do business in Australia.&nbsp;
-                              <a href='https://abr.gov.au/For-Business,-Super-funds---Charities/Applying-for-an-ABN/' target="_blank" rel="external noopener noreferrer">Apply for an ABN here.</a>
-                          </span>)}
-                          readOnly={isNumber(supplierCode)}
-                          disabled={isNumber(supplierCode)}
+                          readOnly={ABN ? true : false}
+                          disabled={ABN ? true : false}
+                          validators={{
+                              validABN: val => val === 'N/A' || validABN(val)
+                          }}
```
ndavison: The desired behaviour here is to disable the ABN field when it is present in the application data, which it should be on new signups since this value is gathered on signup and pre-filled in the initial application data.

There will be a time period where sellers who started their initial applications before this PR will not have had the ABN value pre-filled. If they still haven't input a valid value by the time this PR is merged and deployed, then this code should still allow them to do so (as `ABN` should be falsey). If they did input a value before this PR, then the field will be disabled.

In the event they somehow managed to provide an invalid value before this PR, then they may need manual ops intervention to fix it to finish the application.

apps/marketplace/components/SignupForm/SignupForm.js
```diff
@@ -1,280 +1,389 @@
...
+            If you have any issues, <a href="/contact-us">contact the Marketplace</a>.
+          </p>
+        </React.Fragment>
+      )
+      break
+    default:
+      message = <p>{GENERAL_ERROR}</p>
+      break
+  }
+
```
Reshma-M-A: is this extra line needed?

src/bundles/SellerRegistration/components/BusinessDetailsForm/BusinessDetailsForm.js
```diff
@@ -51,6 +51,7 @@ class BusinessDetailsForm extends BaseForm {
             model,
             returnLink,
             supplierCode,
+            ABN,
```
tobiasbrown, line 54: Should we keep the casing consistent with the model's case?

PR #910, opened 2020-02-11
Mar 3534

Reshma-M-A:
This issue with RFX + Training will be fixed since it was done in this ticket:

MAR-3481: Additional Information: Error validation in contact number typebox is not redCode Review

Just fixing it for ATM - spoke to Clippy and she is happy with that.


malmas18:
there are too many unrelated changes here....

PR #925, opened 2020-03-02
Mar 3977 - Unclear error message - valid date unspecified

Reshma-M-A:
This PR will fix up a few things: 
1.  When a user puts in a date that is in the past, the error message will state **'You must enter a start date that is in the future.'** This will not remove or replace validStartDate function, instead this will only appear when the date is invalid, for instance a leap year.

2. When the form is empty, instead of getting 2 messages that says _'you must enter a valid start date'_ and _'Enter an estimated start date for the opportunity'_, it will only show _'Enter an estimated start date'_ 

The 2nd point wasn't in the ticket but I did it anyway since it didn't make much sense. I updated the UI automation test to reflect this.

### DIFF
--- a/apps/marketplace/components/BuyerSpecialist/BuyerSpecialistTimeframesAndBudgetStage.js
+++ b/apps/marketplace/components/BuyerSpecialist/BuyerSpecialistTimeframesAndBudgetStage.js
@@ -4,18 +4,23 @@ import { connect } from 'react-redux'
 import { Form, actions } from 'react-redux-form'
 import Textfield from 'shared/form/Textfield'
 import formProps from 'shared/form/formPropsSelector'
-import { required, dateIs2DaysInFuture, validDate } from 'marketplace/components/validators'
+import { required, dateIs2DaysInFuture, validDate, dateIsInPast } from 'marketplace/components/validators'
 import AUheadings from '@gov.au/headings/lib/js/react.js'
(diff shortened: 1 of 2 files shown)


apps/marketplace/components/BuyerSpecialist/BuyerSpecialistTimeframesAndBudgetStage.js
```diff
@@ -11,8 +11,10 @@ import DateControl from 'marketplace/components/BuyerBriefFlow/DateControl'
 
 const contractLengthRequired = v => required(v.contractLength)
 const startDateRequired = v => required(v.startDate)
-const startDateIsValid = v => validDate(v.startDate)
-const startDateIs2DaysInFuture = v => !startDateRequired(v) || !startDateIsValid(v) || dateIs2DaysInFuture(v.startDate)
+const startDateIsValid = v => !startDateRequired(v) || validDate(v.startDate, false)
+const startDateIsInPast = v => !startDateRequired(v) || validDate(v.startDate, true) || !validDate(v.startDate, false)
```
tobiasbrown: To follow the pattern of the other functions, does it make sense to redefine this as the following?

```suggestion
const startDateIsInPast = v => !startDateRequired(v) || !startDateIsValid(v) || validDate(v.startDate, true)
```

PR #928, opened 2020-03-12
Must join a team

ndavison:
This PR enables an agency to require their users to be a member of a team before they can perform actions such as creating drafts, editing drafts, publishing, creating work orders, downloading responses etc.

It does this by adding a new column to the agency table, `must_join_team`, which will be a toggle in the admin portal agency edit form. Once enabled, any user not in a team in the agency (including those who signup) will be prompted with a page explaining to the user they must join a team, where they can request to join an existing team in the agency. This sends an email to the chosen team's leads, who can decline the request with a reason, or add the user to their team (using the existing team editing process).


apps/marketplace/components/Teams/Dashboard/TeamsOverview.js
```diff
@@ -30,6 +30,10 @@ export class TeamsOverview extends Component {
 
     return (
       <div className={styles.teams}>
+        <div className={styles.optin}>
```
tobiasbrown, line 33: Did you try to use the margin class from `main.scss`?

src/bundles/ApplicationsAdmin/components/Agency/Agency.js
```diff
@@ -43,6 +43,11 @@ class Agency extends React.Component {
     } else {
       data.whitelisted = false
     }
+    if (data.must_join_team === "on") {
```
tobiasbrown, line 46: Where is this `on` value coming from?

apps/marketplace/components/Teams/Dashboard/TeamsOverview.js
ndavison: I did think about it but the component was already importing its own styles with margins etc so added to what was already there.

src/bundles/ApplicationsAdmin/components/Agency/Agency.js
ndavison: This is the value coming back from the checkbox component (on or off) - it's consistent with the other radios in the same form in the lines above this.

PR #967, opened 2020-11-13
updating error validation message

Reshma-M-A:
Updating the error message when users upload the incorrect file type to **"Incorrect file format. Documents must be in .DOC, .XLS, .PPT or .PDF format.**

### DIFF
--- a/apps/shared/form/FileInput.js
+++ b/apps/shared/form/FileInput.js
@@ -52,6 +52,7 @@ class FileInput extends React.Component {
       <div key={fileField} className="callout-no-margin">
         {this.state.errors && (
           <div className="validation-message">
+            {console.log(this.state.errors)}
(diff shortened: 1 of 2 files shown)


apps/shared/form/FileInput.js
```diff
@@ -52,7 +52,7 @@ class FileInput extends React.Component {
       <div key={fileField} className="callout-no-margin">
         {this.state.errors && (
           <div className="validation-message">
-            <span>There was an error uploading the file: {this.state.errors}</span>
+            <span>Incorrect file format. Documents must be in .DOC, .XLS, .PPT or .PDF format.</span>
```
tobiasbrown, line 55: Will this message make sense if the API throws an error that isn't related to the file extension?

PR #997, opened 2021-08-05
HM-3083: Show invalid CSRF token message

tobiasbrown:
This pull request will make it clearer to users that their CSRF token is invalid and will prompt a page refresh to get a new token.


apps/marketplace/actions/appActions.js
```diff
@@ -85,7 +85,11 @@ export const login = data => (dispatch, getState) => {
     data: JSON.stringify(data)
   }).then(response => {
     if (response.error) {
-      dispatch(setErrorMessage(LOGIN_FAILED))
+      if (response.data && response.data.message && response.data.message.toLowerCase().includes('invalid csrf')) {
```
tobiasbrown, line 88: There are a couple of patterns used to handle errors (see below) but I haven't used them here as it might be better to minimise messages from the API particularly on an action like logging in.

```javascript
if (response.data && response.data.message) {
  dispatch(setErrorMessage(response.data.message))
}
```

Revit (approved):

PR #1000, opened 2021-08-30
HM-3233: Clear CSRF token when invalid

tobiasbrown:
This pull request clears the CSRF token if it's invalid after a login or password reset attempt.


apps/marketplace/actions/resetPasswordActions.js
```diff
@@ -26,6 +26,7 @@ export const sendResetPasswordEmail = values => (dispatch, getState) => {
   }).then(response => {
     if (response.error) {
       if (response.data && response.data.message && response.data.message.toLowerCase().includes('invalid csrf')) {
+        dispatch(setAuthState({ csrfToken: '' }))
```
Revit, line 29: I think we should set to 
dispatch(setAuthState({ isAuthenticated: false, userType: '', csrfToken: '' }))

When it comes to this point , login already failed . We should clear up everything related to user

Revit (approved):

PR #1001, opened 2021-11-03
HM-3487: Use buildpack with Node.js v10

tobiasbrown:
The following error was encountered during deployment of the app:

> Unable to install node: no match found for 10.x in [12.22.5 12.22.6 14.17.5 14.17.6 15.13.0 15.14.0 16.7.0 16.8.0]

This is most likely caused by an update to the Cloud Foundry buildpack which dropped support for version 10.x.


dta-tom (approved):
I can see RC and Alpha Deployment now working fine in CircleCI, front end is working as expected.

PR #1002, opened 2021-11-05
HM-2845: Hide empty desirable criteria

tobiasbrown:
This pull request hides the desirable criteria section in the seller response flow if no desirable criteria exist.  It also renders textfields only if the criterion has been defined (ie. is not an empty string).


dta-tom (approved):
Tested RC vs ALPHA. Code and functionality looks good.

PR #1006, opened 2021-11-23
HM-4135: Adjust display of labour hire licences

tobiasbrown:
This pull request adds checkboxes to toggle the display of labour hire sections on the recruiter stage in the application flow. When a checkbox is not selected after being selected, the labour hire licence data for that state will be reset.

### DIFF
--- a/src/bundles/SellerRegistration/components/RecruiterForm/RecruiterForm.css
+++ b/src/bundles/SellerRegistration/components/RecruiterForm/RecruiterForm.css
@@ -18,3 +18,7 @@
 .expiryDate:last-child {
     margin-bottom: 1rem;
 }
+
+.noBackgroundImage::before {
+    background-image: none !important;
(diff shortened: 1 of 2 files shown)


dta-tom (changes requested):
**RecruiterForm.js**
Upon checking a box, adding license and date data, unchecking box, the data is still sent to the database. As a result, if the form is reloaded, the data is read and the checkboxes are turned back on (which shows line 66 working). Also, if the checkbox is turned back on the data is blanked out.

**To be reviewed:**
OnChange function on AUcheckbox --> May not be reading state value correctly
OnClick  function on AUcheckbox --> may not be setting state OR maybe when it sets the state it doesn't register as an 'onChange' event

dta-tom:
Looks good in Alpha.

dta-tom (approved):

PR #1008, opened 2021-12-09
Add maintenance banner

tobiasbrown:
This pull requests adds a banner to the header to warn users about upcoming maintenance.

Screen Shot 2021-12-09 at 7 00 27 pm


dta-tom (approved):
Looks good

PR #1013, opened 2022-02-07
HM-4374: Reorder components on Who Can Respond stage

dta-tyler:
This pull request reorders components on the "Who Can Respond" stage so the radio list appeared before the category dropdown.
The changes also include:

- extra props for PanelCategorySelect component (description - text displayed underneath field label; helpLabel - text displayed in hyperlink for categories & rates page) 

Screenshot

### DIFF
--- a/apps/marketplace/components/BuyerSpecialist/BuyerSpecialistSelectStage.js
+++ b/apps/marketplace/components/BuyerSpecialist/BuyerSpecialistSelectStage.js
@@ -58,8 +58,6 @@ export class BuyerSpecialistSelectStage extends Component {
   handleSellerCategorySelect(category) {
     if (category !== this.props[this.props.model].sellerCategory) {
       this.props.updateSelectedSellerCategory(category)
-      //this.props.resetOpenTo()
-      //this.props.resetSelectedSellers()
     }
   }
 
@@ -127,7 +125,6 @@ export class BuyerSpecialistSelectStage extends Component {
                 }
               ]}
               messages={{}}
-              onChange={() => this.props.resetSelectedSellers()}
             />
           </div>
           {this.props[this.props.model].openTo === 'all' && (
@@ -211,9 +208,7 @@ const mapStateToProps = (state, props) => ({
 
 const mapDispatchToProps = (dispatch, props) => ({
   countSellers: () => dispatch(countLabourHireSellers()),
-  resetSelectedSellers: () => dispatch(actions.change(`${props.model}.sellers`, {})),
   updateSelectedSellers: sellers => dispatch(actions.change(`${props.model}.sellers`, sellers)),
-  resetOpenTo: () => dispatch(actions.change(`${props.model}.openTo`, '')),
   updateSelectedSellerCategory: category => dispatch(actions.change(`${props.model}.sellerCategory`, category))
 })
 
--- a/apps/marketplace/components/BuyerSpecialist/BuyerSpecialistSelectStage.js
+++ b/apps/marketplace/components/BuyerSpecialist/BuyerSpecialistSelectStage.js
@@ -128,7 +128,7 @@ export class BuyerSpecialistSelectStage extends Component {
             />
           </div>
           {this.props[this.props.model].openTo === 'all' && (
-            <AUpageAlert as="warning">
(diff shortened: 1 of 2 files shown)


apps/marketplace/components/BuyerSpecialist/BuyerSpecialistSelectStage.js
```diff
@@ -13,10 +13,10 @@ import { countLabourHireSellers } from 'marketplace/actions/supplierActions'
 
 import mainStyles from '../../main.scss'
 
-const requiredCategory = v => v.sellerCategory
-const requiredChoice = v => !v.sellerCategory || v.openTo
+const requiredChoice = v => v.openTo
 const atLeastOneSeller = v =>
   !v.openTo || v.openTo === 'all' || (v.openTo === 'selected' && v.sellers && Object.keys(v.sellers).length > 0)
+const requiredCategory = v => v.sellerCategory
```
tobiasbrown, line 19: It looks like this line hasn't changed other than the position in the file.  I'd try and avoid making changes like this in the future as it can make it a bit harder to review incoming changes.

apps/marketplace/components/BuyerSpecialist/BuyerSpecialistSelectStage.js
```diff
@@ -89,9 +87,9 @@ export class BuyerSpecialistSelectStage extends Component {
         model={this.props.model}
         validators={{
           '': {
-            requiredCategory,
             requiredChoice,
-            atLeastOneSeller
+            atLeastOneSeller,
+            requiredCategory
```
tobiasbrown, line 92: This is another example of code that didn't have to change.

apps/marketplace/components/BuyerSpecialist/BuyerSpecialistSelectStage.js
```diff
@@ -109,83 +107,82 @@ export class BuyerSpecialistSelectStage extends Component {
...
         />
-        <PanelCategorySelect
-          id="select-seller"
-          categories={categories}
-          onChange={e => this.handleSellerCategorySelect(e.target.value)}
-          selectedCategory={this.props[this.props.model].sellerCategory}
-          label="Category"
-        />
-        {this.props[this.props.model].sellerCategory && (
+        <div>
```
tobiasbrown, line 110: Is this extra `div` necessary?

apps/marketplace/components/BuyerSpecialist/BuyerSpecialistSelectStage.js
```diff
@@ -109,83 +107,82 @@ export class BuyerSpecialistSelectStage extends Component {
...
+                  label: 'Specific ICT Labour Hire sellers',
+                  value: 'selected'
+                }
+              ]}
+              messages={{}}
+            />
           </div>
-        )}
+          {this.props[this.props.model].openTo === 'all' && (
+            <AUpageAlert as="warning">
```
tobiasbrown, line 131: Do you think we should add some space to the bottom of the page alert to separate it from the category select?
```suggestion
            <AUpageAlert as="warning" className={mainStyles.marginBottom3}>
```

apps/marketplace/components/BuyerSpecialist/BuyerSpecialistSelectStage.js
dta-tyler: True, it didn't need to change. Swapping the order of the validators shows error messages in the same order as the seller & category fields, which is only apparent if the user does not complete either field (likely a very rare scenario).

apps/marketplace/components/BuyerSpecialist/BuyerSpecialistSelectStage.js
dta-tyler: Noted- it does make it more cumbersome comparing changes.

apps/marketplace/components/BuyerSpecialist/BuyerSpecialistSelectStage.js
dta-tyler: It doesn't seem to be, at least not anymore.

tobiasbrown (approved):

PR #1029, opened 2022-04-26
HM-4871: Front page redirection

dta-tyler:
This pull request redirects a logged-in user to the front page (instead of the opportunities page), when no redirect has been specified.


dta-tom (approved):
looks good

PR #1032, opened 2022-05-05
HM-5077: Prevent new seller sign ups and applications

dta-tyler:
This pull request prevents sellers from creating an account or submitting a new application. 

Page alerts are shown on the sign up form for sellers and on the _Continue Application_ page.


dta-tom (approved):
Looks good

PR #1035, opened 2022-05-11
Remove category notice on recruiter form

dta-tyler:
This pull request removes a redundant notice which is displayed when a recruiter seller changes to a different seller type when editing their profile.

notice_message


dta-tom (approved):
Looks good I can update profile without needing to check that box because its not there anymore

